PR #30, opened 2023-06-30
Add templates for wimp example

hoetzsch:
This adds two templates in (cS1, log10(cS2)) for the WIMP example:
  - Nuclear recoil signal of a 50GeV WIMP
  - Electronic recoil background with flat energy spectrum

wimp50gev_template er_template


hoetzsch:
> Nice @hoetzsch One question, are these templates normalized?

Yes, both are normalized to 1.

xzh19980906:
> > Nice @hoetzsch One question, are these templates normalized?
> 
> Yes, both are normalized to 1.

I think it's more sensible to normalize ER template to, say, 200? And 50GeV WIMP to the rate with 1e-45 cm^2 xsec?

hammannr:
I think we agreed now that we would prefer to have the nominal expectation value in the templates and only use an external multiplier as a last resort. Right, @kdund ?

hammannr:
But apart from that it looks great, thanks a lot @hoetzsch !! 😊

xzh19980906:
@hoetzsch FYI, if you wanna normalize the rate to 1e-45 cm2 xsec, you need to
1. let norm_type = 'on_sim'
2. let rate = integral(wimp_spectrum) * exposure

To get the integral of wimp_spectrum, [apt.utils.integrate_midpoint](https://github.com/XENONnT/appletree/blob/5074cb253d465baff07a922b659c01f323efd490/appletree/utils.py#L513) is useful. You need to be careful with the time unit of wimp_spectrum, to be the same as the unit of exposure.

hammannr:
I think for our purposes it would be sufficient to just multiply the histograms with some somewhat sensible expectation values.

xzh19980906:
> I think for our purposes it would be sufficient to just multiply the histograms with some somewhat sensible expectation values.

That's also fine.

hoetzsch:
Updated templates:

- ER normalised to 200 events, WIMP to 10 events
- set empty bins to zero (instead of 1.75e-6)
- increased batch size in sim for production
er_template
wimp50gev_template

hammannr (approved):
Thanks a lot @hoetzsch !! 😊 I gave it a try and everything looked tgood. I’d say you can merge it away! 👍🏼

kdund:
> I think we agreed now that we would prefer to have the nominal expectation value in the templates and only use an external multiplier as a last resort. Right, @kdund ?

I think at least this is the better option.

dachengx:
Hey, @hoetzsch . Sorry for reaching out so late. Where can I find the notebook to reproduce it? Just to make a record.

PR #31, opened 2023-07-10
Init code style checking, pytest, and coverage

dachengx:
Also, add a bug report template.

The bot still complains about the docstring of functions and classes. The beautifying of the docstring will be in another separate PR. It also complains about `assert`, we should rethink about whether raise errors instead, but also in another PR.


.github/workflows/pytest.yml
```diff
@@ -0,0 +1,69 @@
...
+      - name: Checkout repo
+        uses: actions/checkout@v3
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install pytest coverage coveralls
+          pip install -r requirements.txt
+
+      - name: Install Appletree
```
kdund, line 49: is this an oversight?

kdund: perhaps just a copy-pasted comment :)

kdund (approved):
Looks good to me -- need to rename appletree to alea, and then good :)?

.github/workflows/pytest.yml
dachengx: Just fixed.

Issue #43, opened 2023-07-17
Unset default parameters do not give readable warnings


kdund:
**Describe the bug**
If you initialise the gaussian example likelihood but do not set default values of the parameters, the warnings do not clearly point out the issue

**To Reproduce**

```
from alea.examples import gaussian_model
simple_model = gaussian_model.GaussianModel()
simple_model.data = simple_model.generate_data(mu=0,sigma=2)
simple_model.fit()
```
Fails with 
```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In [14], line 4
      2 simple_model = gaussian_model.GaussianModel()
      3 simple_model.data = simple_model.generate_data(mu=0,sigma=2)
----> 4 simple_model.fit()

File ~/Desktop/alea_tests/alea/statistical_model.py:243, in StatisticalModel.fit(self, verbose, **kwargs)
    241 # Make the Minuit object
    242 cost.errordef = Minuit.LIKELIHOOD
--> 243 m = Minuit(
    244     MinuitWrap(cost, s_args=self.parameters.names),
    245     **defaults)
    246 fixed_params = [] if fixed_parameters is None else fixed_parameters
    247 fixed_params += self.parameters.not_fittable

File ~/opt/miniconda3/envs/myenv/lib/python3.8/site-packages/iminuit/minuit.py:625, in Minuit.__init__(***failed resolving arguments***)
    617 self._strategy = MnStrategy(1)
    618 self._fcn = FCN(
    619     fcn,
    620     getattr(fcn, "grad", grad),
    621     array_call,
    622     getattr(fcn, "errordef", 1.0),
    623 )
--> 625 self._init_state = _make_init_state(self._pos2var, start, kwds)
    626 self._values = mutil.ValueView(self)
    627 self._errors = mutil.ErrorView(self)

File ~/opt/miniconda3/envs/myenv/lib/python3.8/site-packages/iminuit/minuit.py:2483, in _make_init_state(pos2var, args, kwds)
   2481 for i, x in enumerate(pos2var):
   2482     val = kwds[x] if kwds else args[i]
-> 2483     err = mutil._guess_initial_step(val)
   2484     state.add(x, val, err)
   2485 return state

File ~/opt/miniconda3/envs/myenv/lib/python3.8/site-packages/iminuit/util.py:1308, in _guess_initial_step(val)
   1307 def _guess_initial_step(val: float) -> float:
[...]

PR #44, opened 2023-07-17
Check if some parameter is not set as guess when fitting

kdund:
See issue #43 -- if you do not set guesses manually, the error message is not easy to parse.

### DIFF
--- a/alea/parameters.py
+++ b/alea/parameters.py
@@ -272,6 +272,10 @@ class Parameters:
             new_val = kwargs.get(name, None)
             if (return_fittable and param.fittable) or (not return_fittable):
                 values[name] = new_val if new_val is not None else param.nominal_value
+        if any(i is None for k, i in values.items()):
+            emptypars = ", ".join([k for k, i in values.items() if i is None])
+            raise AssertionError("All parameters must be set explicitly, or have a nominal value,"
+                                 " encountered for: " + emptypars)
         return values
 
     def __getattr__(self, name: str) -> Parameter:
--- a/alea/statistical_model.py
+++ b/alea/statistical_model.py
@@ -221,7 +221,8 @@ class StatisticalModel:
         defaults = self.parameters(**guesses)
         if any(i is None for k, i in defaults.items()):
             emptypars = ", ".join([k for k, i in defaults.items() if i is None])
-            raise AssertionError("All parameters must be set explicitly, or have a guess for the fit, encountered for: " + emptypars)
+            raise AssertionError("All parameters must be set explicitly, or have a guess for the fit,"
+                                 " encountered for: " + emptypars)
 
         cost = self.make_objective(minus=True, **kwargs)
 
--- a/alea/statistical_model.py
+++ b/alea/statistical_model.py
@@ -219,10 +219,6 @@ class StatisticalModel:
         if not self.parameters.values_in_fit_limits(**guesses):
             raise ValueError("Initial guesses are not within fit limits")
         defaults = self.parameters(**guesses)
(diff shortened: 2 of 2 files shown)


dachengx:
Thanks @kdund . Before fitting, the lack of parameters has already yielded error:

```
from alea.examples import gaussian_model

simple_model = gaussian_model.GaussianModel()
simple_model.parameters.mu.nominal_value = 0.
simple_data = simple_model.generate_data()
```

```
Traceback (most recent call last):
  File "/opt/miniconda3/envs/XENONnT_2023.07.1/lib/python3.9/pdb.py", line 1726, in main                                                   
    pdb._runscript(mainpyfile)
  File "/opt/miniconda3/envs/XENONnT_2023.07.1/lib/python3.9/pdb.py", line 1586, in _runscript                                             
    self.run(statement)
  File "/opt/miniconda3/envs/XENONnT_2023.07.1/lib/python3.9/bdb.py", line 580, in run                                                     
    exec(cmd, globals, locals)
  File "<string>", line 1, in <module>
  File "/home/xudc/tt.py", line 1, in <module>
    from alea.examples import gaussian_model
  File "/home/xudc/alea/alea/statistical_model.py", line 125, in generate_data                                                             
    return self._generate_data(**generate_values)
  File "/home/xudc/alea/alea/examples/gaussian_model.py", line 27, in _generate_data                                                       
    hat_mu = stats.norm(loc=mu, scale=sigma).rvs()
  File "/opt/miniconda3/envs/XENONnT_2023.07.1/lib/python3.9/site-packages/scipy/stats/_distn_infrastructure.py", line 468, in rvs         
    return self.dist.rvs(*self.args, **kwds)
  File "/opt/miniconda3/envs/XENONnT_2023.07.1/lib/python3.9/site-packages/scipy/stats/_distn_infrastructure.py", line 1029, in rvs        
    cond = logical_and(self._argcheck(*args), (scale >= 0))
TypeError: '>=' not supported between instances of 'NoneType' and 'int'
```

I think it is better to check the needed arguments(signatures) of `StatisticalModel._generate_data` and show an error when the needed pars is None. What do you think? Will this violate the degrees of freedom of `_generate_data`? My concern here is that there might be parameters that are arguments of `StatisticalModel._generate_data` but actually not used. So it is not accurate to judge which parameters are 'needed' only given arguments of `StatisticalModel._generate_data`.

dachengx (approved):
Generally looks good to go!

dachengx:
I also think that it is better to handle all parameters:

1. who is needed in which function
2. raise an error when lack of pars
3. ...

just in `Parameters` class. Because in our context once the likelihood is defined, the parameters needed for each function are defined. But of course, we can leave some freedom for debugging.

kdund:
> 

Hi, @dachengx do you mean that the error occurs also when calling generate_data? 
We cannot do code introspection to get the signature, sadly, since we need **kwargs to be the signature for the purposes of the xenon+blueice-from-config likelihood

kdund:
Is it better, you think, to check this in the parameters call function?

dachengx:
> > 
> 
> Hi, @dachengx do you mean that the error occurs also when calling generate_data? We cannot do code introspection to get the signature, sadly, since we need **kwargs to be the signature for the purposes of the xenon+blueice-from-config likelihood

Do you mean this fucntion? https://github.com/XENONnT/alea/blob/257ab91ac09293a2fc5514560f80c13ac34d5369/alea/blueice_extended_model.py#L175-L183

dachengx:
Looks great. I will merge it.

PR #45, opened 2023-07-17
Fix likelihood_names check in statistical_model.store_data to handle unnamed likelihoods

kdund:
Check if likelihood_names exists rather than asking for the (nonexisting) get_likelihood_term_names function. Makes store_data work without setting the data name list.

Example: 
```
from alea.examples import gaussian_model
simple_model = gaussian_model.GaussianModel()
simple_model.data = simple_model.generate_data(mu=0,sigma=2)
simple_model.store_data("simple_data",[simple_model.data])
```
failed before since the store_data function was looking for a NotImplementedError but found a NoAttributeError

### DIFF
--- a/alea/statistical_model.py
+++ b/alea/statistical_model.py
@@ -155,9 +155,9 @@ class StatisticalModel:
         if not defined, it will be ["0", "1", ..., "n-1"]
         """
         if data_name_list is None:
-            try:
-                data_name_list = self.get_likelihood_term_names()
-            except NotImplementedError:
+            if hasattr(self, "likelihood_names"):
+                data_name_list = self.likelihood_names
+            else:
                 data_name_list = ["{:d}".format(i) for i in range(len(data_list[0]))]
 
         kw = {'metadata': metadata} if metadata is not None else dict()


dachengx:
I also think that it is better to handle all parameters:

1. who is needed in which function
2. raise an error when lack of pars
3. ...

just in `Parameters` class. Because in our context once the likelihood is defined, the parameters needed for each function are defined. But of course we can leave some freedom for debugging.

dachengx (commented):
I guess we can just use https://github.com/XENONnT/alea/blob/257ab91ac09293a2fc5514560f80c13ac34d5369/alea/statistical_model.py#L178-L186?

kdund:
> I guess we can just use
> 
> https://github.com/XENONnT/alea/blob/257ab91ac09293a2fc5514560f80c13ac34d5369/alea/statistical_model.py#L178-L186
> ?

But this one is the opposite-- for a certain name it should give you the index, but that relies on the same underlying list of the likelihood names we need for this one.

dachengx (approved):
Thanks @kdund ! Looks good to go!

PR #51, opened 2023-07-21
Simplify file structure

dachengx:
Change the package version to `0.0.0`. 

Separate used and unwed functions in original `utils.py` to `utils.py` and `_utils.py`. 

The inherited unused files and folders are named `_*`.


kdund:
I guess the idea is to have this be the first step towards clearing out now-deprecated binference code, right? So first into a _folder, and then later when replaced, we delete it?

dachengx:
> I guess the idea is to have this be the first step towards clearing out now-deprecated binference code, right? So first into a _folder, and then later when replaced, we delete it?

Yes, exactly.

dachengx:
And also do not use nested folders inside `alea/alea`.

kdund (approved):
Works for me.

PR #61, opened 2023-07-27
Recover examples folder, update file indexing, add notebooks folder

dachengx:
Following private discussion at https://xenonnt.slack.com/archives/C04JRF0AZTP/p1690266817770479.

The main idea is that `GaussianModel` is a well-defined and simple model to show a minimum subclass of `StatisticalModel`.


hammannr (approved):
Looks good and runs! I'd suggest to get rid of legacy code with the same PR and then merge 😊

kdund (approved):
I concur with @hammannr , what we want from the _ - files can be copy-pasted in according to need.

kdund (approved):
yup

PR #65, opened 2023-07-27
Remove pdf_cache folder before pytest

dachengx:
## What does the code in this PR do / what does it improve?

`blueice` makes `pdf_cache` folder in `alea/tests` folder when running test. 

https://github.com/JelleAalbers/blueice/blob/7c10222a13227e78dc7224b1a7e56ff91e4a8043/blueice/source.py#L73

https://github.com/JelleAalbers/blueice/blob/7c10222a13227e78dc7224b1a7e56ff91e4a8043/blueice/source.py#L106-L109

We should delete the folder because we do not want to reuse the cached sources when testing. Otherwise, some classes like `TemplateSource` will not be fully tested, because blueice will just use the cached source. 

## Can you briefly describe how it works?

The pytest fixture allows us to do something before initializing the `TestCase`. 

The configuration of pytest fixture is following https://docs.pytest.org/en/stable/how-to/unittest.html#mixing-pytest-fixtures-into-unittest-testcase-subclasses-using-marks. 

## Can you give a minimal working example (or illustrate with a figure)?

Just as the pytest goes. Or

```
coverage run --source=alea -m pytest -s && coverage report
```

## What are the potential drawbacks of the codes?

_Please include the following if applicable:_
  - [ ] _Update the docstring(s)_
  - [ ] _Update the documentation_
  - [ ] _Tests to check the (new) code is working as desired._
  - [ ] _Does it solve one of the open issues on github?_

### _Notes on testing_
 - _Until the automated tests pass, please mark the PR as a draft._

All _italic_ comments can be removed from this template.


kdund:
The combined template guy is required if we want the safeguard back in and it is useful for mismodelling tests I think it should be kept

dachengx:
> The combined template guy is required if we want the safeguard back in and it is useful for mismodelling tests I think it should be kept

What about `SpectrumTemplateSource`?

kdund:
Less important for current use, but I think we should start using it instead of computing separate signals for each WIMP mass etc :)

kdund:
BTW, can we rename the folder to alea_test_pdf_cache or sth? 
Otherwise, we risk deleting peoples pdf_cache

dachengx:
> BTW, can we rename the folder to alea_test_pdf_cache or sth? Otherwise, we risk deleting peoples pdf_cache

The PR is only deleting `pad_cache` at `alea/tests`.

dachengx:
And also the deletion only happens for `pytest`, not everyone will run pytest locally.

kdund:
ah, ok, yes then no prob.

dachengx:
@kdund Can we go with this?

dachengx:
I do not expect the readthedocs action to be passed because we have no docs yet. We can ignore it when merging.

kdund:
When I run pytest, I typically run it in the alea base folder-- that one is still there with this PR-- not sure how to best deal with this or if ignoring is best..

kdund:
ah, wait-- am I understanding it right that this only deletes the cache _before_ running tests (so not in cleanup?)

dachengx:
> ah, wait-- am I understanding it right that this only deletes the cache _before_ running tests (so not in cleanup?)

Yes. It deletes `pdf_cache` before the test.

dachengx:
> IMO it should then, if possible also delete it after running the tests (you can call it from the TestCase TearDown method: [tearDown()](https://docs.python.org/3/library/unittest.html#unittest.TestCase))

Since not all tests are based on `TestCase`, I updated the pytest fixture.

kdund (approved):
For me this now works-- and I understood the doc failure is sth that will be solved post-merge, right, @dachengx ?

dachengx:
> For me this now works-- and I understood the doc failure is sth that will be solved post-merge, right, @dachengx ?

Yes. It will be fixed in https://github.com/XENONnT/alea/pull/66

PR #66, opened 2023-07-27
Make 0.0.0, initialize documentation structure based on readthedocs, add badges to README

dachengx:
This PR make a structure of docs, but did not add any meaningful contents into it. 

The existence of docs will motivate us to carefully add our docstring. 

After this PR, we have (almost) all structure of components in the repo, along with the package itself:

1. .github
2. tests
3. examples
4. docs


docs/source/reference/.gitignore
```diff
@@ -0,0 +1,4 @@
+# Ignore everything in this directory
+*
+# Except this file
+!.gitignore
```
dachengx, line 4: The idea of this file is from https://stackoverflow.com/a/932982/16515081

dachengx:
@kdund I regretted. In this PR, only the structure of the documentation is added. The content needs a lot more effort and I do not want to delay the release further. You might notice that `HISTORY` is very crowded now: https://github.com/XENONnT/alea/blob/777ef9a17f4dc5af41c9a6bf7818da8a859b61da/HISTORY.md

The good news is that the badges work: https://github.com/XENONnT/alea/blob/777ef9a17f4dc5af41c9a6bf7818da8a859b61da/README.md The DOI and pypi badges need the new release. I hope they look fine after the release.

kdund (approved):
no code all docs. Lets see if it compiles, and fix in prod otherwise (it is Friday, after all :) )

PR #68, opened 2023-07-30
Unify and clean code style and docstring

dachengx:


dachengx:
[CodeFactor](https://www.codefactor.io/repository/github/xenonnt/alea) found an issue: Complex Method

It's currently on:
[alea\model.py:292-360](https://github.com/XENONnT/alea/blob/2e4c06cafc99498438eb563d004a752e88fd1951/alea/model.py#L292-L360)
Commit 2e4c06cafc99498438eb563d004a752e88fd1951

hammannr:
I just had a very brief look. Is there a particular benefit of choosing restructured text docstring? As far as I know, sphynx also supports Google fand NumPy-type docstrings. I generally find reStr text terrible to read and to maintain and if there is no clear point in favor of it I’d much prefer to go to one of the alternative standards.

dachengx:
> I just had a very brief look. Is there a particular benefit of choosing restructured text docstring? As far as I know, sphinx also supports Google fand NumPy-type docstrings. I generally find reStr text terrible to read and to maintain and if there is no clear point in favor of it I’d much prefer to go to one of the alternative standards.

OK, I will go to Google style docstring.

dachengx:
@hammannr I updated the docstring, please check the render at https://alea--68.org.readthedocs.build/en/68/.

alea/model.py
```diff
@@ -16,39 +15,45 @@
...
+        - get_parameter_list
+
+    Attributes:
+        data: data of the model
+        _data: data of the model
+        _confidence_level: confidence level for confidence intervals
+        _confidence_interval_kind: kind of confidence interval to compute
+        parameters: parameters of the model
+        confidence_interval_threshold: threshold for confidence interval
+        is_data_set (bool): True if data is set
```
hammannr, line 45: I think it only makes sense to list here the ones that we want people to use. I.e. remove the ones with an underscore and `is_data_set`.

alea/model.py
```diff
@@ -89,49 +97,53 @@ def _define_parameters(self, parameter_definition):
...
 
         Raises:
             ValueError: If the parameters are not within the fit limits
 
         Returns:
-            Data
+            dict or list: generated data
         """
-        # CAUTION:
-        # This implementation won't allow you to call generate_data by positional arguments.
```
hammannr, line 134: I wouldn't remove this. I think it's good to keep this info there for the future.

alea/model.py
```diff
@@ -187,25 +210,39 @@ def nominal_expectation_values(self):
...
+    def get_likelihood_term_from_name(self, likelihood_name: str) -> Dict:
         """
-        Returns the index of a likelihood term if the likelihood has several names
+        Return the index of a likelihood term if the likelihood has several names
+
+        Args:
+            likelihood_name (str): name of the likelihood term
+
+        Returns:
+            Dict: index of the likelihood term
```
hammannr, line 224: I think this returns an int not a dict

alea/model.py
```diff
@@ -318,19 +376,22 @@ def confidence_interval(
...
-        parameter_interval_bounds: range in which to search for the confidence interval edges.
-            May be specified as:
-            - setting the property "parameter_interval_bounds" for the parameter
-            - passing a list here
+        Uses self.fit to compute confidence intervals for a certain named parameter.
         If the parameter is a rate parameter, and the model has expectation values implemented,
-        the bounds will be interpreted as bounds on the expectation value
-        (so that the range in the fit is parameter_interval_bounds/mus)
-        otherwise the bound is taken as-is.
+        the bounds will be interpreted as bounds on the expectation value,
```
hammannr, line 381: Ok very minor detail but start with large T here.

alea/model.py
```diff
@@ -381,9 +442,19 @@ class MinuitWrap:
     """
     Wrapper for functions to be called by Minuit.
     Initialized with a function f and a Parameters instance.
+
+    Attributes:
+        func: function wrapped
+        s_args (list): parameter names of the model
+        _parameters (dict): parameters of the model
```
hammannr, line 449: parameter names and limits

alea/models/blueice_extended_model.py
```diff
@@ -21,27 +21,31 @@ class BlueiceExtendedModel(StatisticalModel):
...
+        likelihood_names (list): List of likelihood names.
+        data_generators (list): List of data generators for each likelihood term.
+
     Args:
         parameter_definition (dict): A dictionary defining the model parameters.
         likelihood_config (dict): A dictionary defining the likelihood.
+
+    Todo:
+        analysis_space could be inferred from the data
+        (assert that all sources have the same analysis space)
```
hammannr, line 38: Is it very common to add the todos here? We already have the TODOs as issues. I think it's not really needed to also have them here (just more stuff to maintain.. 😄 )

alea/models/blueice_extended_model.py
```diff
@@ -21,27 +21,31 @@ class BlueiceExtendedModel(StatisticalModel):
     This class extends the `StatisticalModel` class and provides methods
     for generating data and computing likelihoods based on blueice.
 
+    Attributes:
+        parameters (Parameters): Parameters object containing the parameters of the model.
+        data (dict): Data of the statistical model.
+        is_data_set (bool): Whether data is set.
+        _likelihood (LogLikelihoodSum): A blueice LogLikelihoodSum instance.
+        likelihood_names (list): List of likelihood names.
+        data_generators (list): List of data generators for each likelihood term.
```
hammannr, line 30: Same as in the base clas.. I think I'd limit this to attributed we want people to use.

alea/models/blueice_extended_model.py
```diff
@@ -58,22 +62,23 @@ def from_config(cls, config_file_path: str) -> "BlueiceExtendedModel":
...
     def data(self, data: dict):
         """
         Overrides default setter. Will also set the data of the blueice ll.
         Data-sets are expected to be in the form of a list of one
-        or more structured arrays-- representing the data-sets of one or more likelihood terms.
+        or more structured arrays representing the data-sets of one or more likelihood terms.
         """
         # iterate through all likelihood terms and set the science data in the blueice ll
-        # last entry in data are the generate_values
+        # except the last entry in data, which are the generate_values
```
hammannr, line 76: with the new implementation it doesn't even have to be the last anymore

alea/models/blueice_extended_model.py
```diff
@@ -21,27 +21,31 @@ class BlueiceExtendedModel(StatisticalModel):
...
-        """
+        """Initializes the statistical model."""
         super().__init__(parameter_definition=parameter_definition)
         self._likelihood = self._build_ll_from_config(likelihood_config)
         self.likelihood_names = [t["name"] for t in likelihood_config["likelihood_terms"]]
         self.likelihood_names.append("ancillary_likelihood")
         self.data_generators = self._build_data_generators()
 
-    # TODO analysis_space could be inferred from the data
-    # (assert that all sources have the same analysis space)
```
hammannr, line 43: Ah I see that you removed these here.. I really like TODOs in the code since you can put it exactly where it belongs. Do you have strong opinions about not doing this?

alea/models/blueice_extended_model.py
```diff
@@ -96,7 +115,6 @@ def get_expectation_values(self, **kwargs) -> dict:
         self_copy = deepcopy(self)
         self_copy.data = self_copy.generate_data()
 
-        # TODO: Make a self.likelihood_temrs dict with the likelihood names as keys and the corresponding likelihood terms as values.
```
hammannr, line 99: 😢

alea/models/blueice_extended_model.py
```diff
@@ -208,11 +246,22 @@ def _generate_data(self, **generate_values) -> dict:
         return data
 
     def _generate_science_data(self, **generate_values) -> dict:
-        science_data = [gen.simulate(**generate_values)
-                        for gen in self.data_generators]
+        """Generate data for all likelihood terms."""
```
hammannr, line 249: for all science data terms -- ancillary is also data

alea/parameters.py
```diff
@@ -7,19 +7,18 @@ class Parameter:
...
+        fittable (bool): Indicates if the parameter is fittable or always fixed.
+        ptype (str): The ptype of the parameter.
+        uncertainty (float or str): The uncertainty of the parameter.
+            If a string, it can be evaluated as a numpy or
+            scipy function to define non-gaussian constraints.
+        relative_uncertainty (bool): Indicates if the uncertainty is relative to the nominal_value.
+        blueice_anchors (list): Anchors for blueice template morphing.
+        fit_limits (tuple): The limits for fitting the parameter.
+        parameter_interval_bounds (tuple): Limits for computing confidence intervals
+        _fit_guess (float): The initial guess for fitting the parameter.
```
hammannr, line 20: For the user it's much more useful to write the property `fit_guess` here instead of the secret attribute I'd say.

alea/simulators.py
```diff
@@ -6,6 +6,7 @@
 import scipy.stats as sps
 import blueice
 import multihist as mh
+from blueice.likelihood import BinnedLogLikelihood, UnbinnedLogLikelihood
```
hammannr, line 9: Do we then even need to import the rest of blueice?

hammannr (commented):
Very nice that you cleaned up and structured our docstrings @dachengx ! Looks really good! I only have a few questions and clarifications mainly on the style.
I checked how it looks on dokuwiki and it seems there is only the structure but not the content. Am I missing somehting?

dachengx:
They are here: https://alea--68.org.readthedocs.build/en/68/py-modindex.html @hammannr

alea/model.py
```diff
@@ -16,39 +15,45 @@
...
+        - a data generation function
+            generate_data(self, parameter_1, parameter_2... parameter_n):
+            A function of the same set of named parameters return a full data set.
+    - Methods that you must implement:
+        - _ll
+        - _generate_data
+    - Methods that you may implement:
+        - get_expectation_values
+    - Methods that already exist here:
+        - ll
```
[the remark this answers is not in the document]
dachengx: Functionally `ll` just call `_ll`, so it does exist.

alea/model.py
dachengx: I think here you mean:
```
This implementation won't allow you to call generate_data by arguments that are not in parameters.
```
?

alea/model.py
dachengx: It follows a line that ends with `,`.

alea/models/blueice_extended_model.py
dachengx: Maybe keep it here and delete it later when it is done.

alea/models/blueice_extended_model.py
dachengx: We can trust the users to have the basic knowledge that the underscore indicates that this is a private attribute. The pros to keep them here is to remind developers of the existence of these attributes. What do you think?

alea/models/blueice_extended_model.py
dachengx: To use the rendered formatted Todo in docs.

hammannr:
> They are here: https://alea--68.org.readthedocs.build/en/68/py-modindex.html @hammannr

Ah very nice! So do we need to make the linking to the things in the sidebar manually? I think this would be very useful!

alea/model.py
```diff
@@ -187,25 +210,39 @@ def nominal_expectation_values(self):
...
             raise NotImplementedError("The attribute likelihood_names is not defined.")
 
     def get_parameter_list(self):
-        """
-        Returns a set of all parameters that the generate_data and likelihood accepts
-        """
+        """Return a set of all parameters that the generate_data and likelihood accepts"""
         return self.parameters.names
 
     def make_objective(self, minus=True, **kwargs):
```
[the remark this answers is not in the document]
hammannr: I think it's not needed for iminuit since we tell it which parameters are fixed it won't call other values.

alea/models/blueice_extended_model.py
hammannr: Yes you're right. Let's leave it like this for now, maybe we can add a reference to the methods that should be used in the docstrings like write that `_generate_data` is the special implementation and should be called as `generate_data` defined in the mother class.

alea/model.py
```diff
@@ -46,10 +47,14 @@ class StatisticalModel:
...
         data: pre-set data of the model
-        parameter_definition (dict or list): definition of the parameters of the model
-        confidence_level (float): confidence level for confidence intervals
-        confidence_interval_kind (str): kind of confidence interval to compute
-        confidence_interval_threshold (Callable[[float], float]): threshold for confidence interval
+        parameter_definition (dict or list, optional (default=None)):
+            definition of the parameters of the model
+        confidence_level (float, optional (default=0.9)):
+            confidence level for confidence intervals
+        confidence_interval_kind (str, optional (default=central)):
```
hammannr, line 54: "central"

alea/model.py
```diff
@@ -372,8 +378,8 @@ def _confidence_interval_checks(
     def confidence_interval(
             self, poi_name: str,
             parameter_interval_bounds: Tuple[float, float] = None,
-            confidence_level: float = None,
-            confidence_interval_kind: str = None,
+            confidence_level: Tuple[float] = None,
+            confidence_interval_kind: Tuple[str] = None,
```
hammannr, line 382: Why did you change these to tuple? I think they should just be float and str or not?

hammannr (commented):
Two minor things and some of the previous discussions are still open but apart from that it looks great -- very nice to have such clean and structured docstrings! 🚀

dachengx:
> > They are here: https://alea--68.org.readthedocs.build/en/68/py-modindex.html @hammannr
> 
> Ah very nice! So do we need to make the linking to the things in the sidebar manually? I think this would be very useful!

Maybe later. This PR only makes docstring. :)

alea/model.py
dachengx: Fixed, I don't know why I changed this.

alea/model.py
dachengx: Fixed.

alea/model.py
dachengx: Got it.

dachengx: Added.

alea/model.py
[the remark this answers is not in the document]
dachengx: Added.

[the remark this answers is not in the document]
hammannr: I think to make things clear we should remove the kwargs now from `make_objective` and when it is called, right?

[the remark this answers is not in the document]
dachengx: Removed.

hammannr (approved):
Looks great to me now -- thank you very much for this contribution @dachengx ! 😊

alea/examples/gaussian_model.py
```diff
@@ -1,32 +1,58 @@
...
 
     def _ll(self, mu=None, sigma=None):
+        """
+        Log-likelihood of the model.
+
+        Args:
+            mu (float, optional (default=None)): mean of the gaussian,
+                if None, the nominal value is used
+            sigma (float, optional (default=None)): standard deviation of the gaussian,
+                if None, the nominal value is used
```
kdund, line 38: should we warn that no nominal values are set here?

hammannr:
Ah, right @kdund we catch it in the Parameters class. Then I think everything is fine or not? Probably I'm just missing your point

alea/examples/gaussian_model.py
kdund: We do catch it-- but it is nice to point out this stumbling block in the documentation, in particular since this is a tutorial model, that fails if you use it naively

alea/simulators.py
```diff
@@ -129,60 +144,15 @@ def simulate(self,
             logging.debug("number of events drawn from Poisson: " + str(n_sources))
             if len(self.ll.base_model.sources) == 1:
                 n_sources = np.array([n_sources])
+
         r_data = np.zeros(np.sum(n_sources), dtype=self.dtype)
         i_write = 0
         for i, n_source in enumerate(n_sources):
-            if 0 < n_source:  #dont generate if 0
+            if n_source > 0:  # dont generate if 0
```
kdund, line 151: This is the wrong way, no? Always better to go smaller->larger

alea/utils.py
```diff
@@ -42,8 +41,7 @@ def adapt_likelihood_config_for_blueice(
 
     Args:
         likelihood_config (dict): likelihood config dict
-        template_folder_list (list): list of possible base folders.
-            Ordered by priority.
+        template_folder_list (list): list of possible base folders. Ordered by priority.
```
kdund, line 44: which order? Try first one first or last one first?

notebooks/placeholder.ipynb
kdund, line 1: While "logPDF" is correct in principle here, I would label it likelihood instead since it is used as a function of mu, rather than the observation.

kdund (changes requested):
Hi! Few minor changes-- I think we should warn that the gaussian model must have nominal values set manually, or fitting will not work (i.e. write it in the docstring) and we should label the likelihood likelihood not logpdf in the example placeholder

alea/model.py
```diff
@@ -189,44 +225,62 @@ def nominal_expectation_values(self):
...
+        """Return a set of all parameters that the generate_data and likelihood accepts"""
         return self.parameters.names
 
-    def make_objective(self, minus=True, **kwargs):
+    def make_objective(self, minus: Optional[bool] = True):
+        """
+        Make a function that can be passed to Minuit
+
+        Args:
+            minus (bool, optional (default=True)): if True, the function is multiplied by -1
```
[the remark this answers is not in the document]
hammannr: Yes, good point -- let's drop it. It's a remnant from blueice, we don't need this flexibility here I think.

alea/simulators.py
kdund: In that case, I want us to ignore WPS309 errors.

kdund (approved):

PR #75, opened 2023-08-10
Example notebooks

### LINKED ISSUE #67: Example Notebooks
Once things settle we should add nice example notebooks and initialize [binder](https://mybinder.org/) to make it easy for people to get started! 😊

hammannr:
This PR adds four example notebooks that hopefully make it easier for people to get started with alea (closes #67).

Since jupyter notebooks are terrible to version control and comment I would suggest that everyone just writes their suggestions in bullet points as a comment to the PR and I'll then upload the updated notebooks 😊 

Here's a brief overview of the notebooks:
## 0. Introduction to alea: a simple Gaussian model
* Define the Gaussian statistical model step by step
* Generate data (for nominal and set values)
* Fit the model
* Learn about initializing via a `parameter_definition` dict

## 1. The BlueiceExtendedModel: Rate and shape parameters
* Initialize the simplified WIMP model from a config file and learn the basics of the model
* Generate toy data and learn about the toy science data
* Learn about Blueice rate and shape parameters

## 2. The BlueiceExtendedModel: Fitting and Confidence Intervals
* Fit the model to toy data (free and constrained fit)
* Learn about the ancillary likelihood term
* Construct the asymptotic confidence interval in the WIMP cross-section

## 3. The BlueiceExtendedModel: Sensitivity Studies
 * Use the `Runner` class to compute the sensitivity band for the hypothetical experiment defined in the simplified model.

## Other changes in this PR
During the writing of the example notebooks I also made these changes:
 * Update the README
 * Change the filenames of the shifted ER templates to make more sense
 * Add the `get_source_name_list` method to the BEM to easier link the source index of a toy data event to the source
 * Add a property `likelihood_list ` to the BEM
 * Define the `__str__` of the `Parameters` class. Now, if you for example do `print(model.parameters)` you get a nice table with all the parameters and the properties/attributes. I find this quite helpful for more complex model definitions.
 * Add the new notebooks to the docs and fill in the installation page.


alea/models/blueice_extended_model.py
```diff
@@ -90,14 +93,34 @@ def data(self, data: dict or list):
...
+    def get_source_name_list(self, likelihood_name: str) -> List:
+        """Return a list of source names for a given likelihood term.
+        The order is the same as used in the `source` column of the data,
+        so this can be used to map the indices provided in the data to a
+        source name.
+
+        Args:
+            likelihood_name (str): Name of the likelihood.
+
+        Returns:
```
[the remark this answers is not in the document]
hammannr: Why did we switch to the capital List, actually @dachengx ?

alea/models/blueice_extended_model.py
```diff
@@ -61,7 +63,8 @@ def from_config(cls, config_file_path: str) -> "BlueiceExtendedModel":
         Returns:
             BlueiceExtendedModel: Statistical model.
         """
-        with open(config_file_path, "r") as f:
+        file_path = get_file_path(config_file_path)
+        with open(file_path, "r") as f:
```
dachengx, line 67: Here you can just use `alea.utils.load_yaml`.

alea/models/blueice_extended_model.py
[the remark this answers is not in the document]
dachengx: Until python3.8 `List` is still preferred: https://stackoverflow.com/questions/39458193/using-list-tuple-etc-from-typing-vs-directly-referring-type-as-list-tuple-etc. But for this one I think you can change it back to list. It depends on https://github.com/JelleAalbers/blueice/blob/7c10222a13227e78dc7224b1a7e56ff91e4a8043/blueice/likelihood.py#L80 which is just `list`.

alea/models/blueice_extended_model.py
dachengx: Similar things also apply to the notebooks.

alea/examples/configs/unbinned_wimp_statistical_model_simple.yaml
```diff
@@ -0,0 +1,77 @@
+parameter_definition:
```
dachengx, line 1: We should also test this file in the `tests` folder.

notebooks/3_sensitivity.ipynb
mdoerenkamp, line 1: I saw that there also was an example runner config. Do you want to add that to the tutorial? Or do you first need submitter functions?

notebooks/3_sensitivity.ipynb
```diff
@@ -0,0 +1,288 @@
...
+      ]
+     },
+     "metadata": {},
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "ref_xsec = 1e-45  # cm^2\n",
+    "uls_xsec = uls * ref_xsec\n",
+    "median_ul = np.quantile(uls_xsec, stats.norm().sf(0))\n",
```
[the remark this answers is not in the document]
hammannr: Oh that's a very nice feature! Though in this example I'd prefer to better make visible what exactly I'm doing.

notebooks/3_sensitivity.ipynb
```diff
@@ -0,0 +1,288 @@
...
+    "\n",
+    "# plot CDF of upper limits\n",
+    "plt.plot(np.sort(uls_xsec), np.linspace(0, 1, len(uls_xsec), endpoint=False),\n",
```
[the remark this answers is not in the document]
hammannr: I prefer the approach where you don't have to sacrifice any resolution. But I agree that the steps would be more representative. I now changed it to plot.step, hope that works for you as well 😊

notebooks/3_sensitivity.ipynb
hammannr: Yes exactly. I think this makes more sense once the submitting part is all figured out. For now I decided not to use it for a tutorial but good catch! 😊

hammannr:
Thanks for having a look @mdoerenkamp and @dachengx ! 🚀 
I now implemented your comments for the notebooks and included the new config to the unittests. I also now renamed `generate_and_write` to `generate_and_store` to make the link to the `store_data` method in the model parent class and changed the named hypothesis`null` to `zero` to avoid confusion with the "null-hypothesis", which can be different. Let me know if you have any other suggestions! 😊

mdoerenkamp (approved):
Looks good! The extra explanations help a lot. I think especially once the documentation is up, it is very clear.
It would be nice to extend this series once submitters are up to also include "workflow", ie how to find threshold, how to loop over masses, etc. But this is more to teach new students and to keep results uniform then actual use of alea.

hammannr:
@mdoerenkamp thanks for reviewing the notebooks! Glad you find them helpful 😊 Yes for sure one can extend this once we have everything together 👌🏼

PR #76, opened 2023-08-10
Set best_fit_args to confidence_interval_args if None

kdund:
Using best_fit_args is mostly relevant for 1D slices of higher-dimensional confidence volumes. 
In most cases, we want it to be set to confidence_interval_args. 
This PR sets it to confidence_interval_args, and makes the docstring more explicit. Also moves it to the last place in the argument list to emphasize that it is a rarer argument. 

What do you think about setting confidence_interval_args via kwargs instead?

### DIFF
--- a/alea/model.py
+++ b/alea/model.py
@@ -396,8 +396,8 @@ class StatisticalModel:
             parameter_interval_bounds: Tuple[float, float] = None,
             confidence_level: float = None,
             confidence_interval_kind: str = None,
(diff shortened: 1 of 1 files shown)


jingqiangye (commented):
Hello team

dachengx (commented):
Thanks @kdund . Looks good!

dachengx (approved):

PR #93, opened 2023-09-28
Change example filename extension

kdund:
We found that some people did not realise that inference_interface was the way to look at files, templates etc. 
In an effort to make it clearer, I thought we might change filenames to .ii.h5 (not mandatory, of course, filenames can be anything) 
Also I added an exclude so the flake8 will not enforce line length in .rst files for example :)


hammannr (commented):
Looks good @kdund ! And the paragraph in the documentation is very helpful I think 😊 .
I found two more occurrences of .h5 in the 3_sensitivity.ipynb notebook and the defaults of the Runner class. It probably makes sense to also replace those.

hammannr (approved):
prfect, thanks! 😊 I'll merge this now so that we can merge #127 afterward

PR #98, opened 2023-10-19
Update requirements.txt

yuema137:
For some reason, it doesn't recognize the blueice installed via ssh url. https is sometimes unstable so ssh is more common for people to use. So I suggest changing the requirement to ssh.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,9 +1,5 @@
 atomicwrites
-{% if USE_HTTPS %}
-git+https://github.com/JelleAalbers/blueice
-{% else %}
 git+ssh://git@github.com/JelleAalbers/blueice.git@main#egg=blueice
-{% endif %}
 h5py
 iminuit
 git+ssh://git@github.com/XENONnT/inference_interface.git@main#egg=inference_interface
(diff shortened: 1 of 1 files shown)


hammannr (dismissed):
Good suggestion! I thing ash is the way to go. Thanks Yue!

hammannr:
Ah I just saw that our CI pipeline doesn't like it. I'll have a look later

hammannr:
Unfortunately, I didn't find a quick fix. I think we could set up deploy keys to solve this but maybe @dachengx has a better idea? Otherwise, I'll try that later!

dachengx:
> Unfortunately, I didn't find a quick fix. I think we could set up deploy keys to solve this but maybe @dachengx has a better idea? Otherwise, I'll try that later!

Usually, ssh authentication requires a valid ssh key, so it is not used for `requirement.txt`. I do not think there can be a quick fix.

hammannr:
After reading a bit about it and trying to find a good solution, I agree @dachengx . @yuema137 maybe this is an issue with the container? Let's discuss this in tech-support and try to find a solution for you rather than changing the requirements

dachengx:
Fixed by https://github.com/XENONnT/alea/pull/132 and https://github.com/JelleAalbers/blueice/pull/42

Issue #81, opened 2023-08-16
Dealing with fixed shape parameters


hammannr:
Some parameters, like the WIMP mass, are typically fixed in our analysis. However, currently, it is possible to change the nominal value of the `wimp_mass` parameter or fix it to a different value in the fit, which just doesn't do anything. I think we should either enable this or raise an error in case the user wants to modify such a parameter value.

Probably the easiest is to introduce a property `static` of a parameter that is true if
- the parameter has no blueice_anchors
- is not a livetime
- is not a rate
If you try to change the value of a static parameter we will throw an error.

I guess in case we want to allow this, we would need to reload the likelihood so you might as well just make a new instance of the model with the new value.

kdund:
Imo we should have static be a default for settable, perhaps just set a "resettable" flag to th relevant ones?

PR #100, opened 2023-10-30
Update Neyman threshold when changing runner_args

hammannr:
For shape parameters such as the WIMP mass, we haven't implemented anything that would allow resetting its nominal value after initialization. However, this is done to obtain the ``poi_rate_multiplier`` for a given ``poi_expectation``, which means that the poi_rate_multiplier values corresponding to a given threshold might be wrong in the end.
I did the following to fix this issue and prevent us from running into a similar issue in the future:
  - Initialize a runner for each runner_arg in the ``NeymanConstructor``. This takes a bit longer and may sometimes be overkill but it works.
  - Introduce a derived boolean property for a parameter called ``static`` (see #81 ), which is true if the parameter is
    - not fittable
    - neither a livetime nor rate (ptype)
    - has no blueice anchors
  - If one tries to override the nominal value of a static parameter or call parameters with a value for it that is different to its nominal value, alea will throw an error.

### DIFF
--- a/alea/model.py
+++ b/alea/model.py
@@ -95,16 +95,19 @@ class StatisticalModel:
         self._confidence_interval_kind = confidence_interval_kind
         self.confidence_interval_threshold = confidence_interval_threshold
         self.asymptotic_dof = asymptotic_dof
-        self._define_parameters(parameter_definition)
+        nominal_values = kwargs.get("nominal_values", {})
+        self._define_parameters(parameter_definition, nominal_values)
 
         self._check_ll_and_generate_data_signature()
-        self.set_nominal_values(**kwargs.get("nominal_values", {}))
 
-    def _define_parameters(self, parameter_definition):
+    def _define_parameters(self, parameter_definition, nominal_values=None):
         """Initialize the parameters of the model."""
         if parameter_definition is None:
             self.parameters = Parameters()
         elif isinstance(parameter_definition, dict):
+            for name, definition in parameter_definition.items():
+                if name in nominal_values:
+                    definition["nominal_value"] = nominal_values[name]
             self.parameters = Parameters.from_config(parameter_definition)
         elif isinstance(parameter_definition, list):
             self.parameters = Parameters.from_list(parameter_definition)
--- a/alea/parameters.py
+++ b/alea/parameters.py
@@ -24,6 +24,8 @@ class Parameter:
             for any value in between.
         fit_limits (Tuple[float, float], optional (default=None)):
             The limits for fitting the parameter.
+        static (bool, optional (default=False)):
+            If True, the parameter is static and cannot be changed from its nominal value.
         parameter_interval_bounds (Tuple[float, float], optional (default=None)):
             Limits for computing confidence intervals.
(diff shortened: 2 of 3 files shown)


hammannr:
I think it would be really neat if the static flag would set itself though I found that this is not really practical for our simple examples and thus I went back to making it a hand-settable flag.

hammannr:
In principle one could also get around loading the runner explicitly by just obtaining the ``poi_expectation`` and ``poi_value`` from ``results[true_name]``, right? I didn't do that now because I wanted to change as little as possible but is there a reason why we should not do that @dachengx ?

alea/model.py
```diff
@@ -95,16 +95,19 @@ def __init__(
         self._confidence_interval_kind = confidence_interval_kind
         self.confidence_interval_threshold = confidence_interval_threshold
         self.asymptotic_dof = asymptotic_dof
-        self._define_parameters(parameter_definition)
+        nominal_values = kwargs.get("nominal_values", {})
+        self._define_parameters(parameter_definition, nominal_values)
 
         self._check_ll_and_generate_data_signature()
-        self.set_nominal_values(**kwargs.get("nominal_values", {}))
```
dachengx, line 101: Is the nearby functionality different from simply using `set_nominal_values`? @hammannr

hammannr: I had to change it since otherwise the static parameters would complain if the nominal value were updated in this step

hammannr: So the outcome is the same. It only had to be changed to accommodate the static parameter functionality

dachengx:
> In principle one could also get around loading the runner explicitly by just obtaining the ``poi_expectation`` and ``poi_value`` from ``results[true_name]``, right? I didn't do that now because I wanted to change as little as possible but is there a reason why we should not do that @dachengx ?

Logically the thresholds json should be controlled by runner configuration yaml but not the files stored. But it is still beneficial to check the consistency between `result` and calculated rate multiplier? 

Better put this in another PR, as an enhancement.

alea/model.py
dachengx: Then maybe `set_nominal_values` is not checked by pytest. Would you add a test to increase the coverage?

hammannr: Actually, now it's no longer in use so I would suggest to simply trash the function

hammannr:
For some reason, pre-commit complains about `_uncertainty` in `parameters`, which I didn't touch:
```
alea/parameters.py:78: error: Cannot determine type of "_uncertainty"  [has-type]
alea/parameters.py:79: error: Cannot determine type of "_uncertainty"  [has-type]
alea/parameters.py:81: error: Cannot determine type of "_uncertainty"  [has-type]
```

dachengx:
> For some reason, pre-commit complains about `_uncertainty` in `parameters`, which I didn't touch:
> 
> ```
> alea/parameters.py:78: error: Cannot determine type of "_uncertainty"  [has-type]
> alea/parameters.py:79: error: Cannot determine type of "_uncertainty"  [has-type]
> alea/parameters.py:81: error: Cannot determine type of "_uncertainty"  [has-type]
> ```

I will try to fix it.

dachengx (approved):

PR #101, opened 2023-10-30
Improve documentation

hammannr:
I started writing some documentation a couple of weeks ago but I can't find the time to finish it the next time.

However, I think the things that are added are already helpful for beginners. For this reason, I would suggest merging it at this stage and then we can pick it up from this point the next time someone has time to write something.


hammannr:
Do we really need doc8 to annoy us if the line is too long? I thought we discussed this at some point and anyway one would use alt+z in vscode or something for writing doc

hammannr:
To check the compiled version you can have a look [here](https://alea--101.org.readthedocs.build/en/101/) 😊

kdund:
> Do we really need doc8 to annoy us if the line is too long? I thought we discussed this at some point and anyway one would use alt+z in vscode or something for writing doc

Agreed, if possible I would remove all linting of docfiles.

kdund (dismissed):
I agree with merging

kdund (approved):

PR #103, opened 2023-11-02
Only initialize runner once in NeymanConstructor

dachengx:
This is a further fix of https://github.com/XENONnT/alea/pull/100. We do not need to initialize runner for each `**{**nominal_values, **generate_values}`, but we can pass `generate_values, nominal_values` to `runner.update_poi` to pass correct `nominal_values` to get the expectations. Additionally, I made modifications:
1. Require parameter's `ptype` to be in `["rate", "shape", "efficiency", "livetime"]`, the default `ptype` is `"shape"`.
2. Recover `StatisticalModel.set_nominal_values` and `Parameters.set_nominal_values`, add option to modify the static parameter so that user is aware of the modification.

### DIFF
--- a/alea/examples/configs/unbinned_wimp_statistical_model_simple.yaml
+++ b/alea/examples/configs/unbinned_wimp_statistical_model_simple.yaml
@@ -1,7 +1,10 @@
 parameter_definition:
   wimp_mass:
     nominal_value: 50
+    ptype: shape
     fittable: false
+    blueice_anchors:
+      - 50
     description: WIMP mass in GeV/c^2
 
   livetime:
--- a/alea/model.py
+++ b/alea/model.py
@@ -95,19 +95,16 @@ class StatisticalModel:
         self._confidence_interval_kind = confidence_interval_kind
         self.confidence_interval_threshold = confidence_interval_threshold
         self.asymptotic_dof = asymptotic_dof
-        nominal_values = kwargs.get("nominal_values", {})
-        self._define_parameters(parameter_definition, nominal_values)
+        self._define_parameters(parameter_definition)
 
         self._check_ll_and_generate_data_signature()
+        self.set_nominal_values(overwrite_static=True, **kwargs.get("nominal_values", {}))
 
-    def _define_parameters(self, parameter_definition, nominal_values=None):
+    def _define_parameters(self, parameter_definition):
         """Initialize the parameters of the model."""
         if parameter_definition is None:
             self.parameters = Parameters()
         elif isinstance(parameter_definition, dict):
-            for name, definition in parameter_definition.items():
-                if name in nominal_values:
-                    definition["nominal_value"] = nominal_values[name]
             self.parameters = Parameters.from_config(parameter_definition)
         elif isinstance(parameter_definition, list):
             self.parameters = Parameters.from_list(parameter_definition)
@@ -238,6 +235,18 @@ class StatisticalModel:
         if len(_data_list[0]) != len(data_name_list):
             raise ValueError("The number of data sets and data names must be the same")
         toydata_to_file(file_name, _data_list, data_name_list, **kw)
+
+    def set_nominal_values(self, overwrite_static=False, **nominal_values):
+        """Set the nominal values for parameters.
+
+        Keyword Args:
+            nominal_values (dict): A dict of parameter names and values.
+
+        """
+        self.parameters.set_nominal_values(
+            overwrite_static=overwrite_static,
+            **nominal_values,
+        )
 
     def set_fit_guesses(self, **fit_guesses):
         """Set the fit guesses for parameters.
--- a/alea/parameters.py
+++ b/alea/parameters.py
@@ -13,7 +13,7 @@ class Parameter:
         nominal_value (float, optional (default=None)): The nominal value of the parameter.
         fittable (bool, optional (default=True)):
             Indicates if the parameter is fittable or always fixed.
-        ptype (str, optional (default=None)): The ptype of the parameter.
+        ptype (str, optional (default=shape)): The ptype of the parameter.
         uncertainty (float or str, optional (default=None)): The uncertainty of the parameter.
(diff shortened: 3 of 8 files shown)


alea/submitters/local.py
```diff
@@ -187,15 +191,22 @@ def submit(
...
-            runner_args["statistical_model_args"].pop("limit_threshold", None)
-            runner = Runner(**runner_args)
-            expectation_values = runner.model.get_expectation_values(
-                **{**nominal_values, **generate_values}
-            )
-            # in some rare cases the poi is not a rate multiplier
-            # then the poi_expectation is not in the nominal_expectation_values
-            component = self.poi.replace("_rate_multiplier", "")
-            poi_expectation = expectation_values.get(component, None)
+            if runner.input_poi_expectation:
```
kdund, line 194: Why not just "if poi_expectation" in generate_values?

dachengx: Should also be fine. What is the benefit?

alea/submitters/local.py
```diff
@@ -187,15 +191,22 @@ def submit(
...
+                poi_expectation = generate_values.get("poi_expectation")
+                # nominal_values are passed to update_poi to update the poi
+                # like wimp_mass, livetime, etc.
+                generate_values = runner.update_poi(
+                    runner.model, self.poi, generate_values, nominal_values
```
hammannr, line 199: I think this won't work with static parameters. That was the entire point of the previous PR

hammannr: I ran it quickly and yes, this crashes if you have static parameters (as it should).

alea/examples/configs/unbinned_wimp_statistical_model_simple.yaml
```diff
@@ -1,7 +1,10 @@
 parameter_definition:
   wimp_mass:
     nominal_value: 50
+    ptype: shape
     fittable: false
+    blueice_anchors:
+      - 50
     description: WIMP mass in GeV/c^2
```
dachengx, line 8: This modification is to be compatible with the requirement of `ptype`.

alea/model.py
```diff
@@ -95,19 +95,16 @@ def __init__(
...
         self.asymptotic_dof = asymptotic_dof
-        nominal_values = kwargs.get("nominal_values", {})
-        self._define_parameters(parameter_definition, nominal_values)
+        self._define_parameters(parameter_definition)
 
         self._check_ll_and_generate_data_signature()
+        self.set_nominal_values(overwrite_static=True, **kwargs.get("nominal_values", {}))
 
-    def _define_parameters(self, parameter_definition, nominal_values=None):
+    def _define_parameters(self, parameter_definition):
```
dachengx, line 103: I think `_define_parameters` should be as simple as just initializing `Parameters`.

alea/model.py
```diff
@@ -239,6 +236,18 @@ def store_data(
             raise ValueError("The number of data sets and data names must be the same")
         toydata_to_file(file_name, _data_list, data_name_list, **kw)
 
+    def set_nominal_values(self, overwrite_static=False, **nominal_values):
```
dachengx, line 239: Recover `set_nominal_values`, with an option to overwrite static parameters. And in principle I think we do need a method to overwrite static parameters(bypass the check).

alea/parameters.py
```diff
@@ -54,6 +54,11 @@ def __init__(
         self.name = name
         self._nominal_value = nominal_value
         self.fittable = fittable
+        if ptype not in ["rate", "shape", "efficiency", "livetime"]:
```
dachengx, line 57: Here the requirement of `ptype` is more strict.

alea/submitters/local.py
dachengx: The recent commit 7b4f38fe8c4bbb9d1dfb41ba95638deaa3a32040 makes the PR compatible with static parameters.

alea/submitters/local.py
```diff
@@ -226,7 +238,7 @@ def submit(
                         "threshold": [],
                         "poi_expectation": [],
                     }
-                    threshold[threshold_key] = threshold_value
+                    threshold[threshold_key] = deepcopy(threshold_value)
```
dachengx, line 241: Very important because the shallow copy will convey the modification of `nominal_values` and `generate_values` to `threshold_value` and `threshold`.

alea/utils.py
```diff
@@ -449,7 +449,7 @@ def convert_variations(variations: dict, iteration) -> list:
         if not isinstance(v, list):
             raise ValueError(f"variations {k} must be a list, not {v} with {type(v)}")
         variations[k] = expand_grid_dict(v)
-    result = [dict(zip(variations, t)) for t in iteration(*variations.values())]
+    result = [dict(zip(variations, deepcopy(t))) for t in iteration(*variations.values())]
```
dachengx, line 452: Very important because the shallow copy will convey the modification of items in `t` in one entry of `result`, to another entry of `result`.

alea/model.py
```diff
@@ -95,19 +95,16 @@ def __init__(
         self._confidence_interval_kind = confidence_interval_kind
         self.confidence_interval_threshold = confidence_interval_threshold
         self.asymptotic_dof = asymptotic_dof
-        nominal_values = kwargs.get("nominal_values", {})
-        self._define_parameters(parameter_definition, nominal_values)
+        self._define_parameters(parameter_definition)
 
         self._check_ll_and_generate_data_signature()
+        self.set_nominal_values(overwrite_static=True, **kwargs.get("nominal_values", {}))
```
dachengx, line 101: @hammannr Can we keep `set_nominal_values ` in the future so that `_define_parameters` only handle the definition of `Parameters`? And for the `need_reinit` parameters, I think they can be defined correctly just by `set_nominal_values`.

hammannr:
This PR can be closed now, right @dachengx ?

dachengx:
Suspended by https://github.com/XENONnT/alea/pull/108

PR #104, opened 2023-11-06
Print time usage of `Runner.run`

dachengx:
Just print real and CPU time.


hammannr (approved):
Looks good! I think it's good to have that at least for now while we're still developing the code!

PR #106, opened 2023-11-08
Get expectation values per likelihood term

hammannr:
This PR adds additional functionality to the `get_expectation_values` method in the `BlueiceExtendedModel`. By default, the expectation values for the same source are summed over all likelihood terms. This remains to be the default, however now you can also get a dict of a dict, which contains the expectation values for each lieklihood term individually.

### Example:

```python
model = BlueiceExtendedModel.from_config("unbinned_wimp_statistical_model.yaml")

print(model.get_expectation_values())
# out: {'wimp': 12.0, 'er': 440.0}

print(model.get_expectation_values(per_likelihood_term=True))
# {'sr0': {'er': 40.0, 'wimp': 2.0}, 'sr1': {'er': 400.0, 'wimp': 10.0}}

```

### DIFF
--- a/alea/models/blueice_extended_model.py
+++ b/alea/models/blueice_extended_model.py
@@ -194,11 +194,10 @@ class BlueiceExtendedModel(StatisticalModel):
                 ret[ll_name][n] = mu
         if not per_likelihood_term:
             # sum over sources with same names of all likelihood terms
-            all_source_names = [ret[ll_name].keys() for ll_name in ret.keys()]
-            all_source_names = set().union(*all_source_names)
+            all_source_names = {name for sublist in ret.values() for name in sublist}
             ret = {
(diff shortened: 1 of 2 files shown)


hammannr:
The pre-commit doesn't like the way I change names I think. But I kind of prefer changing the type of ret instead of having to returns in this function.. What do you think?

kdund:
To me it looks fine-- mypy is not happy though (I think the overloaded type of ret?

kdund:
Any way you can ask mypy to skip those lines?

hammannr:
Actually, I think the cleanest way would be to always return a dict of dicts and in the default case it only has one key "total". But I kind of hate to make the default more cumbersome to use... So I think I'll just overload this and make mypy ignore the line.

dachengx (approved):

PR #107, opened 2023-11-09
Prevent arguments to submission variations being changed by deepcopy-ing them. 

dachengx:
Without deepcopy, the dictionary inside returned list of function `convert_variations` will be entangled:

```
>>> convert_to_vary({"a": [1, 2], "b": [{"c": 3}, {"c": 4}]}) #  referred as `result`
[{'a': 1, 'b': {'c': 3}}, {'a': 1, 'b': {'c': 4}}, {'a': 2, 'b': {'c': 3}}, {'a': 2, 'b': {'c': 4}}]
```

When you modify the `result[0]['b']`, without this PR, `result[2]['b']` will also be changed, because only the address is shallowly copied.

This PR is split from https://github.com/XENONnT/alea/pull/103.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -12,6 +12,7 @@ from alea.utils import (
     clip_limits,
     can_expand_grid,
     expand_grid_dict,
+    convert_to_vary,
     deterministic_hash,
 )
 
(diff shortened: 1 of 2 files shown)


tests/test_utils.py
```diff
@@ -74,6 +75,21 @@ def test_expand_grid_dict(self):
             ],
         )
 
+    def test_convert_to_vary(self):
+        """Test of the convert_to_zip function."""
+        varied = convert_to_vary({"a": [1, 2], "b": [{"c": 3}, {"c": 4}]})
+        self.assertNotEqual(id(varied[0]["b"]), id(varied[2]["b"]))
```
kdund, line 81: This only checks the basic sanity and not the issue where if you change one you dont change the other, right? I wonder if I would change this to "
```
btest = dict(c=4)
varied = convert_to_vary({"a": [1, 2], "b": [{"c": 3}, btest]})
btest["c"] = 5
self assertEqual(varied[1]["b"], dict(c=4))
```

?

kdund (approved):
Thanks, @dachengx this protects the careless :) 
If you think it makes sense, it would be good to add to the test case like I suggest, I think

tests/test_utils.py
dachengx: `self.assertNotEqual(id(` is checking the address of dictionaries, not just a sanity check.

kdund: you're right, thanks, @dachengx

PR #108, opened 2023-11-10
Read poi and expectation directly from `output_filename` to accelerate `NeymanConstructor`

dachengx:
We do not need to initialize runner for each **{**nominal_values, **generate_values}, but we can read the poi directly from toymc files `output_filename`.

This was initially an idea from @hammannr, https://github.com/XENONnT/alea/pull/100#issuecomment-1784817837.

0. Save expectation_values to metadata
1. First figure out whether it is single `output_filename` mode or multiple(batch) `output_filename` mode
2. Read metadata from the list of `output_filename` and make sure that the metadata is the same in the list
3. Take the first entry of the metadata list as metadata
4. Assert it contains poi, which is guaranteed by https://github.com/XENONnT/alea/blob/80bd030205ed299fd26aedaddfdc31f3ee5849bb/alea/runner.py#L340
5. Read poi and expectation_values from metadata
6. Read poi_expectation from expectation_values

This PR is superseding https://github.com/XENONnT/alea/pull/103.

### DIFF
--- a/alea/models/blueice_extended_model.py
+++ b/alea/models/blueice_extended_model.py
@@ -2,6 +2,7 @@ import warnings
 from typing import List, Dict, Callable, Optional, Union, cast
 from copy import deepcopy
 from pydoc import locate
+import itertools
 
 import numpy as np
 import scipy.stats as stats
@@ -138,10 +139,30 @@ class BlueiceExtendedModel(StatisticalModel):
         ll_index = self.likelihood_names.index(likelihood_name)
         return self.likelihood_list[ll_index].source_name_list
 
+    def get_all_source_name(self) -> set:
+        """Return a set of possible source names from all likelihood terms.
+
+        Args:
+            likelihood_name (str): Name of the likelihood.
+
+        Returns:
+            set: set of source names.
+
+        """
+        source_name = set(
+            itertools.chain.from_iterable([ll.source_name_list for ll in self.likelihood_list[:-1]])
+        )
+        return source_name
+
     @property
     def likelihood_list(self) -> List:
         """Return a list of likelihood terms."""
         return self._likelihood.likelihood_list
+
+    @property
+    def likelihood_parameters(self) -> List:
+        """Return a list of likelihood parameters."""
+        return self._likelihood.likelihood_parameters
 
     def get_expectation_values(self, per_likelihood_term=False, **kwargs) -> dict:
         """Return total expectation values (summed over all likelihood terms with the same name)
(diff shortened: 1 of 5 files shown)


alea/models/blueice_extended_model.py
```diff
@@ -138,11 +139,31 @@ def get_source_name_list(self, likelihood_name: str) -> list:
...
+
+        Returns:
+            set: set of source names.
+
+        """
+        source_name = set(
+            itertools.chain.from_iterable([ll.source_name_list for ll in self.likelihood_list[:-1]])
+        )
+        return source_name
+
```
hammannr, line 156: This is not used anywhere. I would remove it

alea/models/blueice_extended_model.py
```diff
@@ -138,11 +139,31 @@ def get_source_name_list(self, likelihood_name: str) -> list:
...
     @property
     def likelihood_list(self) -> List:
         """Return a list of likelihood terms."""
         return self._likelihood.likelihood_list
 
+    @property
+    def likelihood_parameters(self) -> List:
+        """Return a list of likelihood parameters."""
+        return self._likelihood.likelihood_parameters
+
```
hammannr, line 166: I don't mind it but also this is not related to this PR, is it?

alea/models/blueice_extended_model.py
hammannr: also, it doesn't take `likelihood_name` and I think if at all it should be `get_all_source_names`.

alea/submitters/local.py
```diff
@@ -156,19 +158,82 @@ def submit(
...
+                else:
+                    warnings.warn(
+                        "Can not find the poi_expectation in the generate_values "
+                        f"{generate_values}, so will not check if the poi_expectation "
+                        "are consistent!"
+                    )
+                    poi_expectation = _poi_expectation
+
+            # read the likelihood ratio
+            results = toyfiles_to_numpy(output_filename)
```
hammannr, line 236: Can we test some of this?

hammannr: In general, this method has become pretty lengthy. Do you think it could make sense to break out some functionality into smaller functions that you can then call in `submit`? This makes it easier to test and to read 😊

hammannr (commented):
I didn't go through it line-by-line yet but I think it looks good!
I already added a few comments and I'm now running a small toy study to see if everything works as expected. I'll let you know as soon as it's done 😊

alea/models/blueice_extended_model.py
dachengx: Suspended by https://github.com/XENONnT/alea/pull/111

alea/submitters/local.py
dachengx: The functions are split. I have no idea currently how to test the code of `NeymanConstructor` because then we need to save the output files or run the configs when testing. That can be time consuming.

hammannr: You're right. Maybe we can write and load some trivial example or something? But it's not that straightforward, I agree.

hammannr (approved):
The split version is now much more readable -- I like it! I think we should think about unittests for the `NeymanConstructor` but I'd suggest to merge this now without them.

I ran the code with some examples and the results were sensible. Thanks a lot @dachengx for this improvement 😊

Issue #105, opened 2023-11-08
Improve "static" functionality


hammannr:
After a discussion with @kdund and @dachengx we decided to improve the static functionality introduced in #100 in the following way:
  - instead of having a property `static` we make it a `ptype`, which we'll call `needs_reinit`
  - This ptype will be the fallback option in the blueice extended model in case nothing else will be specified
  -  For `wimp_mass` e.g. it would make sense to use this parameter because for most applications you won't use it as a shape parameter but then you can't be allowed to change it's nominal value without re-initializing the entire model.

kdund:
I think this issue has now been met/accomplished? (& well done for it!)

PR #110, opened 2023-11-13
Make sure values of parameters that need re-initialization are not changed

hammannr:
As mentioned in #105 this PR improves the implementation of parameters that require the model to be re-initialized (such as the wimp_mass in case it is not treated as shape parameter.

For this, I added the `ptype` `"needs_reinit"`, which sets a related boolean property. The behaviour is then the same as before with the `static` boolean.

In the blueice extended model we now check that all ptypes are known and in case none are provided we fall back to needs_reinit.

### DIFF
--- a/alea/examples/configs/unbinned_wimp_statistical_model.yaml
+++ b/alea/examples/configs/unbinned_wimp_statistical_model.yaml
@@ -1,7 +1,6 @@
 parameter_definition:
   wimp_mass:
     nominal_value: 50
-    ptype: shape
     fittable: false
     description: WIMP mass in GeV/c^2
     blueice_anchors:
--- a/alea/examples/configs/unbinned_wimp_statistical_model.yaml
+++ b/alea/examples/configs/unbinned_wimp_statistical_model.yaml
@@ -3,8 +3,6 @@ parameter_definition:
     nominal_value: 50
     fittable: false
     description: WIMP mass in GeV/c^2
-    blueice_anchors:
-      - 50
 
   livetime_sr0:
     nominal_value: 0.2
--- a/alea/models/blueice_extended_model.py
+++ b/alea/models/blueice_extended_model.py
@@ -438,7 +438,7 @@ class BlueiceExtendedModel(StatisticalModel):
         If no ptype is specified, set the default ptype "needs_reinit".
 
         """
-        allowed_ptypes = ["rate", "shape", "efficiency", "livetim", "needs_reinit"]
+        allowed_ptypes = ["rate", "shape", "efficiency", "livetime", "needs_reinit"]
         default_ptype = "needs_reinit"
         for p in self.parameters:
             if p.ptype is None:
--- a/alea/models/blueice_extended_model.py
+++ b/alea/models/blueice_extended_model.py
@@ -60,6 +60,7 @@ class BlueiceExtendedModel(StatisticalModel):
         ]
         self.livetime_parameter_names += [None]  # ancillary likelihood
         self.data_generators = self._build_data_generators()
+        self._set_default_ptype()
 
     @classmethod
     def from_config(cls, config_file_path: str, **kwargs) -> "BlueiceExtendedModel":
@@ -430,6 +431,23 @@ class BlueiceExtendedModel(StatisticalModel):
                 " must be constrained to be finite"
             )
         ll.add_shape_parameter(efficiency_name, anchors=(limits[0], limits[1]))
+
+    def _set_default_ptype(self):
+        """Check if all parameters have a ptype that is in the list of allowed ptypes.
+
+        If no ptype is specified, set the default ptype "needs_reinit".
+
+        """
+        allowed_ptypes = ["rate", "shape", "efficiency", "livetim", "needs_reinit"]
+        default_ptype = "needs_reinit"
+        for p in self.parameters:
+            if p.ptype is None:
+                p.ptype = default_ptype
+            elif p.ptype not in allowed_ptypes:
+                raise ValueError(
+                    f"Parameter {p.name} has ptype {p.ptype} which is not in the list of "
+                    f"allowed ptypes: {allowed_ptypes}."
+                )
 
     def store_real_data(self, file_name: str, real_data_list: list, metadata=None):
         """Store real data in a file with toydata format.
--- a/tests/test_blueice_extended_model.py
+++ b/tests/test_blueice_extended_model.py
@@ -203,3 +203,8 @@ class TestBlueiceExtendedModel(TestCase):
             # check that a ValueError is raised when calling the parameters with another value
             with self.assertRaises(ValueError):
                 model.parameters(wimp_mass=1)
+            with self.assertRaises(ValueError):
+                model.get_expectation_values(wimp_mass=1)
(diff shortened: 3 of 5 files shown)


alea/models/blueice_extended_model.py
```diff
@@ -431,6 +432,23 @@ def _set_efficiency(self, source: dict, ll):
...
+        default_ptype = "needs_reinit"
+        for p in self.parameters:
+            if p.ptype is None:
+                p.ptype = default_ptype
+            elif p.ptype not in allowed_ptypes:
+                raise ValueError(
+                    f"Parameter {p.name} has ptype {p.ptype} which is not in the list of "
+                    f"allowed ptypes: {allowed_ptypes}."
+                )
+
```
dachengx, line 451: I think you just need to set a default value at https://github.com/XENONnT/alea/blob/bd71e948d7a57191bd1fd28e750fe2e439c6f8c8/alea/parameters.py#L43 to `ptype`, and maybe also do any kind of test there.

If the `needs_reinit` is just for `BlueiceExtendedModel`, I think you should change the `parameter_definition ` at https://github.com/XENONnT/alea/pull/110/files#diff-ccb14fbc2f6819c7774476fe82d0c97563c7006391df64c86b0fbea0706af8f5L52 then. Because it is not a good practice to change the definition of `ptype` outside parameters.py.

dachengx (changes requested):
In general I think we should revert the decision on https://github.com/XENONnT/alea/blob/bd71e948d7a57191bd1fd28e750fe2e439c6f8c8/alea/model.py#L103 and https://github.com/XENONnT/alea/commit/684deae8044f7c72b50498e006ff9e0627fe277f. The `set_nominal_values` should be kept and inside `_define_parameters` we should not change anything in the config.

Also, another more important functionality is when users run `get_expectation_values` and `_ll` or other functions whose result is relevant to the `needed_reinit` parameters, they should be warned or prevented the usage of `needed_reinit` parameters in `kwargs`.

dachengx:
Would you please update the title so that it is more catchy? @hammannr

hammannr:
> Would you please update the title so that it is more catchy? @hammannr

Catchy enough @dachengx or do you want a Nature Physics title? 😄

dachengx:
> Catchy enough @dachengx or do you want a Nature Physics title? 😄

Good, good, good enough.

kdund:
"A novel method to prevent catastrophic mistakes when discovering datk matter"

hammannr:
> In general I think we should revert the decision on
> 
> https://github.com/XENONnT/alea/blob/bd71e948d7a57191bd1fd28e750fe2e439c6f8c8/alea/model.py#L103
> 
> and [684deae](https://github.com/XENONnT/alea/commit/684deae8044f7c72b50498e006ff9e0627fe277f). The `set_nominal_values` should be kept and inside `_define_parameters` we should not change anything in the config.
> Also, another more important functionality is when users run `get_expectation_values` and `_ll` or other functions whose result is relevant to the `needed_reinit` parameters, they should be warned or prevented the usage of `needed_reinit` parameters in `kwargs`.

I think the `set_nominal_values` is not really compatible with `needs_reinit` because the parameter is first set to the default and then it will complain when it's changed (even though the model is not yet fully initialized). Still, I feel like making a workaround for that is more cumbersome than just setting the nominal values to the dictionary before initializing the parameters. This also limits the misuse of overriding nominal values of a parameter, which doesn't like those kinds of things...

For the second part, I think I can't follow you. Right now, whenever you try to set a value other than the nominal it will simply raise an error.

alea/models/blueice_extended_model.py
hammannr: We don't want this to be the default. This means that you _always_ have to specify a ptype basically. The Gaussian Model for example would become more complex than it needs to be and this then also translates to any other likelihoods added in the future. I think it's ok to have functionality in parameters that you can take into your project if you like but don't have to.

alea/examples/configs/unbinned_wimp_statistical_model.yaml
```diff
@@ -1,7 +1,6 @@
 parameter_definition:
   wimp_mass:
     nominal_value: 50
-    ptype: shape
     fittable: false
     description: WIMP mass in GeV/c^2
     blueice_anchors:
```
dachengx, line 6: The anchors here should be removed. We should also raise an error if anchors are provided for `needs_reinit` parameter.

tests/test_blueice_extended_model.py
```diff
@@ -192,3 +192,14 @@ def test_store_real_data(self):
...
+    def test_needs_reinit(self):
+        """Test of the needs_reinit property of parameters."""
+        for model in self.models:
+            self.assertTrue(model.parameters["wimp_mass"].needs_reinit)
+            # check that a ValueError is raised when trying to set the nominal parameter
+            with self.assertRaises(ValueError):
+                model.parameters["wimp_mass"].nominal_value = 1
+            # check that a ValueError is raised when calling the parameters with another value
+            with self.assertRaises(ValueError):
+                model.parameters(wimp_mass=1)
```
dachengx, line 205: We should also add test like:
```
            # check that a ValueError is raised when calling the expectation values with another value
            with self.assertRaises(ValueError):
                model.get_expectation_values(wimp_mass=1)
            # check that a ValueError is raised when calling the fit with another value
            with self.assertRaises(ValueError):
                model.fit(wimp_mass=1)
```

dachengx (changes requested):
I have no more confusion about my second point.

But for the first point, for example, the `Parameters.from_list` called by `_define_parameters` should also accept modification of `nominal_values`. So for completeness, we should update the `nominal_values` also when `parameter_definition` is a list. But this workaround is still cumbersome.

Generally, it is better to separate fictionality into different functions. I made a PR about this part: https://github.com/XENONnT/alea/pull/114

alea/examples/configs/unbinned_wimp_statistical_model.yaml
hammannr: removed -- good point. Not sure if we need to raise an error though. Maybe rather a warning if you pass anchors for a parameter that's not a shape? Because doing this won't hurt anyone it's just useless, right?

dachengx: Yes, you are right. My point is only to be strict. Maybe just a warning, because anyway we raise error in `fit` and `get_expectation_values` if another value is assigned.

hammannr:
Thanks, @dachengx ! Though I'm not convinced that set_nominal is the better option here. I'm afraid that making this option so accessible can lead to misuse. I would simply add the missing implementation for the parameter_list and use the current implementation.

dachengx:
> Thanks, @dachengx ! Though I'm not convinced that set_nominal is the better option here. I'm afraid that making this option so accessible can lead to misuse. I would simply add the missing implementation for the parameter_list and use the current implementation.

People will be curious about `overwrite_init`. I don't think it is so dangerous.

We can benefit from an odd number of workers. What is your @kdund  opinion about https://github.com/XENONnT/alea/pull/114?

hammannr:
@kdund opinions on this? 😊

dachengx (approved):

PR #111, opened 2023-11-13
Add function to get all sources names from all likelihoods

dachengx:
Add function to get all sources names from all likelihoods in `BlueiceExtendedModel`

### DIFF
--- a/tests/test_blueice_extended_model.py
+++ b/tests/test_blueice_extended_model.py
@@ -70,13 +70,13 @@ class TestBlueiceExtendedModel(TestCase):
             expectation_values.append(this_expectation_dict)
         return expectation_values
 
-    def test_get_all_source_names(self):
-        """Test of the get_all_source_names method."""
+    def test_all_source_names(self):
+        """Test of the all_source_names method."""
         for config, model in zip(self.configs, self.models):
             _source_names = set()
             for ll_t in config["likelihood_config"]["likelihood_terms"]:
                 _source_names.update([s["name"] for s in ll_t["sources"]])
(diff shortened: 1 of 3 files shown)


tests/test_blueice_extended_model.py
```diff
@@ -70,6 +70,11 @@ def get_expectation_values(self):
             expectation_values.append(this_expectation_dict)
         return expectation_values
 
+    def test_get_all_source_names(self):
+        """Test of the get_all_source_names method."""
+        for model in self.models:
+            model.get_all_source_names()
+
```
hammannr, line 77: Not sure if checking whether it doesn't crash is a sensible test 😅 Maybe load the config and check manually or something like that? Otherwise this is a bit pointless or not?

dachengx: Added the consistency check in pytest.

hammannr:
Looks good @dachengx !
Should we use it [here](https://github.com/XENONnT/alea/blob/f756e0a77efbf0dd2900e7f830d5741523db2e5a/alea/models/blueice_extended_model.py#L213)?

dachengx:
> Looks good @dachengx ! Should we use it [here](https://github.com/XENONnT/alea/blob/f756e0a77efbf0dd2900e7f830d5741523db2e5a/alea/models/blueice_extended_model.py#L213)?

I set the `all_source_names` as a property of `BlueiceExtendedModel`, and used it in `BlueiceExtendedModel.get_expectation_values` as you suggested.

hammannr (approved):
perfect! Merge away @dachengx ! 😊

Issue #115, opened 2023-11-14
Allow arbitrary computation names


kdund:
https://github.com/XENONnT/alea/blob/bcad6fdaf8b53767360d327f52ad179310c46646/bin/alea-submission#L13

It can be useful to just add some throw-away computations in addition to the main named ones (            "discovery_power", "threshold", "sensitivity", opinions? It should just be a question of throwing a warning rather than an error if another name is used as long as it exists in the run config file.

hammannr:
But if it's defined in the config but not in the submitter/runner it can't do anything sensible or not? Do you maybe have an example of where/how you would like to use it? 😊

kdund:
Sure, lets say you want to do a small set of toyMCs to check something, perhaps you want to make a pull-plot-- you copy one of the current dict entries, change the generate_args and fit parameters to your liking, and then you wish to submit those-- the submission is exactly the same as for "sensitivity"-- what changes is that you perhaps only ask for the "true" fit, and you don't compute confidence intervals. 

Or at least to my knowledge, there is nothing that changes in how toyMCs are run between "discovery power" and "sensitivity"-- it is just the content of the run config dict entry

hammannr:
You're probably right! The only thing which is for sure different is the threshold computation. Then this sounds good to me!

PR #116, opened 2023-11-15
Allow all computation names

kdund:
#115 -- the world is rich, and we can allow any computation name (though "threshold" is special, and others such as "collate results" may be in the future)

### DIFF
--- a/bin/alea-submission
+++ b/bin/alea-submission
@@ -10,12 +10,26 @@ def main():
         "--computation",
         type=str,
         required=True,
-        choices=[
-            "discovery_power",
-            "threshold",
-            "sensitivity",
-        ],
-        help="Type of computation, defined in YAML runner config file",
+        help=(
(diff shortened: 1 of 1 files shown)


bin/alea-submission
```diff
@@ -10,12 +10,26 @@ def main():
         "--computation",
         type=str,
         required=True,
-        choices=[
-            "discovery_power",
-            "threshold",
-            "sensitivity",
-        ],
-        help="Type of computation, defined in YAML runner config file",
+        help=(
```
kdund, line 13: my IDE complains about """  block quotes, but the solution here seems way worse for readability surely?

hammannr: Ah got it. You want the names to have "" -- then I think this is fine.

hammannr (approved):
Hey @kdund sorry for the delay. Thanks for liberating the computation names!! 😊 Looks good!

PR #122, opened 2023-12-14
Optional setting of random seed for debugging

dachengx:
Provide `seed` as a keyword argument of `Runner`, so that we can reproduce the result when debugging because it is deterministic. In productive usage, it might not be necessary.

### DIFF
--- a/alea/runner.py
+++ b/alea/runner.py
@@ -66,6 +66,7 @@ class Runner:
         toydata_filename (str, optional (default=None)): toydata filename
         only_toydata (bool, optional (default=False)): whether only generate toydata
         output_filename (str, optional (default='test_toymc.h5')): output filename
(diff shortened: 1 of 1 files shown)


kdund:
I think this would also only be feasible if np.random is the _only_ module used for random numbers, right? Not sure if that is true at the moment?

dachengx:
> I think this would also only be feasible if np.random is the _only_ module used for random numbers, right? Not sure if that is true at the moment?

Yes, I also had this concern. But `scipy` also uses seed same as `np.random`. So I think it is true for alea.

hammannr (commented):
I ran some tests with the `unbinned_wimp_statistical_model_simple` and everything went as expected. But if @kdund has concerns/suggestions for improvement I'll leave the approval of this PR to him.

kdund:
If just for testing, in fine with this then:)

PR #123, opened 2023-12-14
Save dtype of `valid_fit` as bool

dachengx:
Because it is bool. Previously the `valid_fit` is a float.


kdund (approved):
One typo, otherwise approved

PR #124, opened 2023-12-16
Evaluate `blueice_anchors` expression

dachengx:
Try to evaluate `blueice_anchors` if it is a string.


hammannr (dismissed):
Looks good and works! When playing around with it a bit I was considering to suggest treating this similar to the uncertainty property such that you can also change it via a string after initialization. I still haven't completely made up my mind about it, in principle `blueice_anchors` shouldn't be changed much so I think the way you implemented it is ok. But if you think it's nicer to implement the "translation functionality" via a property, feel free to do so! 😊

dachengx:
> Looks good and works! When playing around with it a bit I was considering to suggest treating this similar to the uncertainty property such that you can also change it via a string after initialization. I still haven't completely made up my mind about it, in principle `blueice_anchors` shouldn't be changed much so I think the way you implemented it is ok. But if you think it's nicer to implement the "translation functionality" via a property, feel free to do so! 😊

I think your suggestion is rather mandatory for me. Sorry, I forget the setter function. I think the setter is more elegant here and we should always use such elegant solutions.

PR #126, opened 2023-12-21
Tiny minor change on docstring

dachengx:


docs/source/reference/modules.rst
dachengx, line 1: This is not needed.

hammannr (approved):
Looks good, thanks! 😊

PR #127, opened 2024-01-04
Add axis_names to example templates

hammannr:
While doing the tutorials @SphericalCowww spotted that the example templates don't have axis names! Adding them makes it easier for people to get started.


SphericalCowww:
I just checked, and I think it works out nicely for me :D

SphericalCowww (approved):
Looks great to me.

PR #128, opened 2024-01-07
Improve check of already made toydata and output

dachengx:
Previously we only checked whether `output_filename` was made regardless of `only_toydata`, and did not checked `toydata_filename`.

In this PR, we check if:

1. `output_filename` is made when not `only_toydata`
2. `toydata_filename` is made when `toydata_mode` is `generate_and_store`


hammannr (approved):
Looks good, thanks for adding this! 😊

PR #131, opened 2024-01-13
Combine several jobs into one to save computation resources

dachengx:
Develop a new ticket generator: `combined_tickets_generator` which combines `combine_n_job` jobs into one by `" && "`.

Because the initialization of the container and environment will take some time, which is a constant offset proportional to the number of jobs. So this PR will reduce those time.

### DIFF
--- a/alea/submitter.py
+++ b/alea/submitter.py
@@ -317,7 +317,7 @@ class Submitter:
         """Get the submission script for the current configuration. It generates the submission
         script for each combination of the computation options.
 
-        for Runner from to_zip, to_vary and in_common.
+        For Runner from to_zip, to_vary and in_common:
             - First, generate the combined computational options directly.
             - Second, update the input and output folder of the options.
             - Thrid, collect the non-fittable(settable) parameters into nominal_values.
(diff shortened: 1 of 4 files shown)


hammannr:
@dachengx I quickly went over the code -- just for my understanding: Isn't this the same as just reducing the `n_batch` parameter?

dachengx:
@hammannr I see your concern. Sorry, I was not clear enough in the description. Each job will have a unique `nominal_value` or `generate_value` for a parameter. Sometimes the number of combinations of different `nominal_value`s and `generate_value`s is very large. If I want to do a gridscan of two parameters' nominal_value`s, and each has 100 different choices, I will need to submit 10000 jobs. This PR mainly solves such a problem. If I set `combine_n_jobs` to 100, then I will only need to submit 100 jobs.

alea/submitter.py
```diff
@@ -387,6 +388,31 @@ def already_done(self, i_args: dict) -> bool:
             is_done = False
         return is_done
 
+    def combined_tickets_generator(self):
+        """Get the combined submission script for the current configuration. ``self.vcombine_n_job``
```
hammannr, line 392:
```suggestion
        """Get the combined submission script for the current configuration. ``self.combine_n_job``
```

hammannr: could you maybe add a small example to make clear how/why one should use the combine_n_jobs? 😊

hammannr (dismissed):
Looks good! Two minor comments then you can go ahead and merge I'd say 😊

PR #135, opened 2024-01-22
Update `hypotheses` and `common_hypothesis` by `pre_process_poi`

dachengx:
Previously only the `generate_values` are updated by `update_poi`:

https://github.com/XENONnT/alea/blob/cee1e71858fa967ecb00f123dc43d0c9b4127d91/alea/runner.py#L172-L181

This PR also updates `"poi_expectation"` in `hypotheses` and `common_hypothesis`.

### DIFF
--- a/alea/runner.py
+++ b/alea/runner.py
@@ -184,8 +184,7 @@ class Runner:
         else:
             self.input_poi_expectation = False
         # update poi according to poi_expectation
-        value = self.pre_process_poi(value, "generate_values")
-        self._generate_values = value
+        self._generate_values = self.pre_process_poi(value, "generate_values")
 
     @property
     def common_hypothesis(self) -> Dict[str, float]:
@@ -194,8 +193,7 @@ class Runner:
     @common_hypothesis.setter
     def common_hypothesis(self, value: Dict[str, float]) -> None:
         # update poi according to poi_expectation
-        value = self.pre_process_poi(value, "common_hypothesis")
-        self._common_hypothesis = value
+        self._common_hypothesis = self.pre_process_poi(value, "common_hypothesis")
 
     @property
     def hypotheses(self) -> list:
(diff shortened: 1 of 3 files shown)


alea/runner.py
```diff
@@ -164,33 +164,48 @@ def __init__(
             statistical_model_args.get("asymptotic_dof", 1),
         )
 
+    def pre_process_poi(self, value, name):
+        """Pre-process of poi_expectation."""
+        if not all([isinstance(v, (float, int)) for v in value.values()]):
+            raise ValueError(f"{name} should be a dict of float! But {value} is provided.")
+        # update poi according to poi_expectation
+        if "poi_expectation" in value:
+            value = self.update_poi(self.model, self.poi, value, self.nominal_values)
```
hammannr, line 173: Since this is the only place where you use `update_poi` why don't you just make it a regular method (non-static) and then only parse `value` here?

alea/runner.py
```diff
@@ -164,33 +164,48 @@ def __init__(
             statistical_model_args.get("asymptotic_dof", 1),
         )
 
+    def pre_process_poi(self, value, name):
```
hammannr, line 167: `name` is a bit generic. Maybe something like `called_by_name` instead to make it a bit clearer? But I can also live with name :D

hammannr (approved):
Thanks a lot @dachengx for this quick implementation! I tried it with a few toys and it works flawlessly. I left two minor comments that you can implement if you find them useful, otherwise just go ahead and merge 😊

alea/runner.py
dachengx: Maybe users will need it in their notebook.

alea/runner.py
dachengx: I updated it to `attribute_name`.

alea/runner.py
hammannr: good point!

PR #142, opened 2024-02-27
Make "piecewise" the default pdf interpolation

hammannr:
This PR makes `"piecewise"` the default pdf interpolation method. Before, the default was defined in blueice to `"linear"`, which is only in certain limits compatible with the `BlueiceDataGenerator`. The `BlueiceDataGenerator` is now only assigned if all sources have piecewise interpolation -- in the future we can add a suitable generator also for the linear interpolation case.

### DIFF
--- a/alea/models/blueice_extended_model.py
+++ b/alea/models/blueice_extended_model.py
@@ -347,7 +347,23 @@ class BlueiceExtendedModel(StatisticalModel):
 
         """
         # last one is AncillaryLikelihood
-        return [BlueiceDataGenerator(ll_term) for ll_term in self.likelihood_list[:-1]]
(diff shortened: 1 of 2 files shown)


hammannr:
@kdund you could add `pdf_interpolation_method="linear"` in one of the example configs and try to load the model -- it should complain in the initialization step that it doesn't like it. If you add `pdf_interpolation_method="piecewise"` or nothing at all it should work fine.

kdund (approved):
Checked, and works, thanks @hammannr

dachengx:
Hey @hammannr . Maybe you could try to squash and merge next time. Then it is easier to track the change from a PR in the GitHub history.

PR #143, opened 2024-02-27
point away from alea for physics models

kdund:
Several look to alea to understand our signal or background model, this adds a short text to point out that alea takes this from otherwhere


hammannr (approved):
Nice text! Thanks 👍

PR #144, opened 2024-02-28
Enforce bins in config and template to match

hammannr:
So far, we only strictly enforced the lenghts of bin edges in config and template to match. I would suggest enforcing this also if values are off. In some cases the bin volumes are relevant and thus overriding the edges with wrong values might cause problems.


dachengx (approved):
Looks good. Thanks @hammannr .

PR #146, opened 2024-03-15
Make local submitter verbose

hammannr:
This small PR simply adds some useful information output when submitting with `--local --debug`.
In particular, it will print the parameter overview table and if available the nominal expectation values.


dachengx (approved):
LGTM Thanks @hammannr

PR #149, opened 2024-03-18
Consistent sorting for BlueiceExtendedModel

hammannr:
@dachengx noticed that there is an inconsistency between the sorting when using `get_expectation_values()` and `get_source_histograms(...)`. This is not critical since only dictionaries are returned but can lead to confusion when looping through items.

This PR fixes this issue and sorts all returns by the source name.


dachengx (changes requested):
Thanks @hammannr . Please add a test.

hammannr:
With the example configs, this test is not super powerful but it will check if we screw something up badly in the future at least. I found it a bit of an overkill to add a complicated model just to test this functionality but what do you think @dachengx ?

PR #151, opened 2024-03-20
Print Argument combinations to be submitted

hammannr:
This is just a very small change for submission in debug mode.
It will print all parameter combinations that would be submitted, which can be helpful when using many `to_zip` and `to_vary`arguments.


dachengx (approved):
Thanks @hammannr . LGTM

PR #152, opened 2024-03-21
Fixed data storing

hammannr:
If toy data is already assigned to the model, it will be converted to `alea.utils.ReadOnlyDict`. So far, this type was not supported by the builtin `model.store_data()` method, which is changed with this PR.

Now you can:

```python
model = ...
model.store_data(file_name, [model.data])
```


dachengx (changes requested):
Thanks @hammannr . Looks good to me. Please add a test to make sure that the previously ignored usage case is supported after this PR.

dachengx (approved):

PR #153, opened 2024-03-22
Add midway2/3 partition/qos

shenyangshi:
A small change that makes the automatic differentiation between midway3 and midway2.

### DIFF
--- a/alea/submitters/slurm.py
+++ b/alea/submitters/slurm.py
@@ -30,7 +30,7 @@ class SubmitterSlurm(Submitter):
         self._eval_partition()
         # suggested default arguments for utilix.batchq.submit_job
         # for XENONnT collaboration
-        BATCHQ_DEFAULT_ARGUMENTS = {
+        batchq_default_arguments = {
             "hours": 1,  # in the unit of hours
             "mem_per_cpu": 2000,  # in the unit of Mb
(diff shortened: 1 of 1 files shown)


alea/submitters/slurm.py
```diff
@@ -39,11 +26,37 @@ def __init__(self, *args, **kwargs):
...
         self.log_dir = self.outputfolder
 
+    def _eval_partition(self):
+        # automatically set default partition based on hostname
+        hostname = socket.gethostname()
+        if "midway3" in hostname:
+            self.partition = "lgrandi"
+            self.qos = "lgrandi"
+            self.exclude_nodes = ""
+        else:
```
dachengx, line 55: If people run on other servers, not RCCm, they will be confused by this condition.

dachengx (changes requested):
Maybe we should:
1. add a condition for midway2 and midway3
2. warn the user if the hostname is not understandable and assign nothing

shenyangshi:
excluded_nodes as `null` is added as default for midway3, as without an explicit excluded_nodes, midway3 job submission will yield error.

hammannr:
@dachengx after #238 things have changed. Does it make sense to redo this PR or should we close it?

PR #156, opened 2024-04-08
Fitting index variables

zihaoxu98:
This PR modifies the default model.fit to support mixing continuous variables and index variables, with **little** extra time consuming. In this PR I only considered index variables as the shape parameter, i.e. a new ptype called `shape_index`. But the function can be easily used whenever we need other index variables in the future.

I did this because shape parameter usually causes many fitting problems since the likelihood can hardly be convex within the fitting limit if the shape parameter is not defined in a very careful way. Thus an alternative method is to do grid search on anchors as well as the `migrad()`. However if you simply do `migrad()` on all anchor points the computation speed will be significantly lowered. In this PR I do it in this way:

  - Fix index variables to their guesses, fit other normal parameters.
  - Fix other normal parameters to the best fit in last step, grid search index variables and find the best fit index.
  - Iterate the above two steps for some times until converge.

In most of the cases we care about, this iteration usually just need to be done once to reach convergence.

Here is an example JIC you wanna have a quick test,
```
import numpy as np
import matplotlib.pyplot as plt

np.random.seed(1)

config = "unbinned_wimp_statistical_model_index_fitting.yaml"
model = BlueiceExtendedModel.from_config(config)
model.fit()
m = model.minuit_object

er_band_shift = model.parameters.parameters['er_band_shift'].blueice_anchors
ll = [model.ll(er_band_shift=x) for x in er_band_shift]

plt.scatter(er_band_shift, ll)
plt.axvline(m.values['er_band_shift'], label='model.fit')
plt.legend()
plt.xticks(er_band_shift)
plt.xlabel('er_band_shift as index variable')
plt.ylabel('model.ll')
plt.show()
```
Screenshot 2024-04-08 at 8 23 59 PM

### DIFF
--- a/alea/examples/configs/unbinned_wimp_statistical_model_index_fitting.yaml
+++ b/alea/examples/configs/unbinned_wimp_statistical_model_index_fitting.yaml
@@ -34,7 +34,7 @@ parameter_definition:
 
   er_band_shift:
     nominal_value: 0
-    ptype: shape_index
+    ptype: index
     uncertainty: null
     fittable: true
     blueice_anchors:
--- a/alea/examples/configs/unbinned_wimp_statistical_model_index_fitting.yaml
+++ b/alea/examples/configs/unbinned_wimp_statistical_model_index_fitting.yaml
@@ -54,6 +54,7 @@ likelihood_config:
     - name: science_run
       default_source_class: alea.template_source.TemplateSource
       likelihood_type: blueice.likelihood.UnbinnedLogLikelihood
+      likelihood_config: {"morpher": "IndexMorpher"}
       analysis_space:
         - cs1: np.arange(0, 102, 2)
         - cs2: np.geomspace(100, 100000, 51)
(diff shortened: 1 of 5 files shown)


alea/examples/configs/unbinned_wimp_statistical_model_index_fitting.yaml
```diff
@@ -0,0 +1,79 @@
...
+    relative_uncertainty: true
+    fittable: true
+    fit_limits:
+      - 0
+      - null
+    fit_guess: 1.0
+
+  er_band_shift:
+    nominal_value: 0
+    ptype: shape_index
```
dachengx, line 37: Better name it as `index`.

zihaoxu98: Then we have to assume `index` will always be the shape parameter?

dachengx: What is the other kind of index you will have?

zihaoxu98: idk, if ppl think it's fine I am ok...

dachengx (approved):
Thanks @zihaoxu98 . Looks cool.

PR #161, opened 2024-05-01
Add sources and example notebook for 1D CES inference

yuema137:
## Introduction
This PR introduces two new fitting capabilities to Alea: unbinned and binned likelihood fitting for robust 1D inference. These additions expand Alea's functionality without modifying existing behavior.

### Framework Updates
The following general framework modifications improve flexibility while maintaining backward compatibility:

1. Enhanced `_prefix_file_path` in `utils.py` to support both single filenames and filename lists, previously limited to single strings only

2. Fixed configuration initialization in `blueice_extended_model` to properly handle `source_wise_interpolation` from user inputs
3. Pass the `fiducial_mass` parameter in `blueice_extended_model` to the sources
4. Improved logging in `simulator` by adding per-source expected event counts for better observability

The changes are focused on generalizing functionality and improving logging, with no breaking changes to existing features.

### 1D Inference Implementation
This is a refactored version of the [inference developed for SR0](https://github.com/XENONnT/nton/tree/master/nton/lower). Featured improvements:

- Accommodate to common alea format to take advantage of existing alea functions
- **Simplify the structure**: Instead of having multiple layers in nton.lower, here we only need to add a specific transformation and source to the ces. The rest of the function is handled by existing bases in alea, which users don't need to touch (in nton.lower users need to modify all of the layers in order to add a background or change transformation). 
- **Decouple transformations (smearing, bias, efficiency) from sources**. Therefore users can feed their self-defined transformations into the framework without touching the sources.
- **Enable the fitting for the parameters in transformations**: in nton.lower the smearing and energy bias were hard coded and fixed during the fitting. Now they can also be fit if needed.

#### ces_source.py:
- Defines the basic sources needed for 1D CES
- Validates the transformation parameters and models 
- Users can write extended sources if there's a need. But most of time, the basic ones should be enough.

#### ces_transformation.py:
- Defines the basic transformations needed for 1D CES
- Users can add transformations/models if there's a need. Those functions may need to be changed frequently because of the update of data/cuts/corrections.

#### xe133_template.ii.h5
An example template (Hist1d)

#### unblind_ces_simple.yaml:
An example yaml config including:
- A Hist-based component (Xe133)
- A Monoenergetic component (test_gaussian)
- A Flat component (test_flat)

#### 4_ces_inference.ipynb
An example notebook showing how to build model based on example ymal config and do fittings

#### 5_ces_binned.ipynb
Similar to 4, but with binned likelihood

#### Efficiency
A key design choice in the 1D inference implementation concerns the handling of efficiency corrections. The implementation bypasses the efficiency application step in the blueice extended model, as the term "efficiency" is reserved and should not be used for 1D sources.

This design decision addresses a fundamental limitation: in most cases, efficiency is not a simple scalar value but rather an energy-dependent curve, typically modeled using empirical functions with shaping parameters. The total efficiency loss depends on these shaping parameters. However, the current blueice framework [treats efficiency as a scalar value per source](https://github.com/JelleAalbers/blueice/blob/0c8adb4639475a3ef7b44f9c5a742eb469fce51b/blueice/likelihood.py#L385-L393).

To resolve this limitation, we adopt the approach used in the [SR0 inference](https://github.com/XENONnT/nton/tree/master/nton/lower) by Jingqiang. The efficiency corrections are applied during source initialization, meaning:

1. The `pdf_histogram` and `expected_events` are pre-corrected for efficiency
2. No further efficiency corrections are needed in subsequent processing
3. The original efficiency application step in the blueice likelihood class is bypassed

This approach provides a more accurate handling of energy-dependent efficiency corrections while maintaining compatibility with the existing framework.
#### Fiducial Mass
As mentioned in the general framework update, we passed the `fiducial_mass_parameter` into the sources. This parameter is similar to the `livetime` parameter, but **it's only called within the ces source class to normalize the event rate**. So, this change should not affect any of the existing functions.


hammannr:
Hey @yuema137 , I started reading the code but it seems there are bunch of flake8 formats of code that is unchanged. Could you please revert those changes? This way we can better review the changes to the code of this PR. If you have strong reason to change some format it would be better to make a separate PR for it 😊

hammannr:
Hey @yuema137 after having a first brief look I have a few comments / suggestions:
- I think it would be really great if we could generalize this for any 1D inference instead of specifically focusing on CES. As far as I can tell there are only a few CES-specific features that I think could be solved differently but please let me know if you think this is not feasible.
- Enable `apply_slice_args` in your source -- very useful feature! 😊 
- Every public method should get at least a short docstring.
- Use piecewise interpolation for now since we don't have a data generator (yet) for anything else (see #142 ).
- The example notebook deserves a bit of text, maybe have a look at the other example notebooks.
- I think we should not generally normalize to events/keV/year.

Now I also have a question for @kdund :
I'm afraid that we run into problems here with the duplication of code in the source and the simulator. So far this works but I think we need to better separate the two. Right now, I think the data generator doesn't use the normalization defined by Yue, right? If I remember correctly, the simulator was built to generate data more efficiently, right? But should we maybe rewrite it such that it calls the simulate method of the source under the hood? This ensures that the correctly normalised/transformed pdf is used, right?

kdund:
Agree that this is very nice and should be our standard 1D inference framework!
(So renaming from ces to oned_model would be a good step)

kdund:
@hammannr to be specific-- you are saying that the alea/simulators.py reuses/duplicates some source .simulate functionality, right? 
I agree that it is cleanest with the sources handling interpolation etc. to the extent possible.

hammannr:
> @hammannr to be specific-- you are saying that the alea/simulators.py reuses/duplicates some source .simulate functionality, right? I agree that it is cleanest with the sources handling interpolation etc. to the extent possible.

I think so, yes. I can check again and try to improve this. But if I don't miss anything, I think the simulate method of the source is currently completely circumvented.

kdund:
The one

> > @hammannr to be specific-- you are saying that the alea/simulators.py reuses/duplicates some source .simulate functionality, right? I agree that it is cleanest with the sources handling interpolation etc. to the extent possible.
> 
> I think so, yes. I can check again and try to improve this. But if I don't miss anything, I think the simulate method of the source is currently completely circumvented.

The one problem I might see with that is that we want the source classes to be fully fledged bluesky sources also so that part kinda has to stay?

hammannr:
> The one
> 
> > > @hammannr to be specific-- you are saying that the alea/simulators.py reuses/duplicates some source .simulate functionality, right? I agree that it is cleanest with the sources handling interpolation etc. to the extent possible.
> > 
> > 
> > I think so, yes. I can check again and try to improve this. But if I don't miss anything, I think the simulate method of the source is currently completely circumvented.
> 
> The one problem I might see with that is that we want the source classes to be fully fledged bluesky sources also so that part kinda has to stay?

Yes -- my idea is to try to use as much of the source functionality as possible in the simulator. At least right now I don't see the reason (yet) why so much of the functionality of the sources is rewritten in the simulator.

yuema137:
Hi @hammannr thanks for the comments! I have done the following tasks:

- Add docstrings to the public methods
- Revert the formatting commit
- Add more explanations to the example notebook

And the following's are ongoing as I need to check if they affect any functions:
- Use piecewise interpolation for now 
- Enable `apply_slice_args`
I think we should not generally normalize to events/keV/year.

For those two points I'm not perfectly sure so further comments would be helpful:
- generalized 1D framework: from my view, this is already a generalized framework as the transformations could be overwritten by users. And they could apply any possible transformations to 1d histogram. So the main change would just be renaming the files and classes. Am I missing anything here?
- Normalization: The motivation to normalize them into event/year/keV but not event/year/ton/keV is that some of the backgrounds don't scale up with the fiducial mass. In the lower inference, we used to calculate rate multipliers in the inference, which made the code hard to modify.  So, I think it's better to separate the statistical inference from the rate multiplier calculation. The rate multiplier should come from external calculations (YAML configs). But if there's a better solution I'm glad to implement

alea/ces_source.py
```diff
@@ -0,0 +1,234 @@
...
+        h = Hist1d(
+            data=np.linspace(self.min_e, self.max_e, number_of_bins),
+            bins=number_of_bins,
+            range=(self.min_e, self.max_e),
+        )
+        h.histogram = h.histogram.astype(np.float64)
+        return h
+
+    def _transform_histogram(self, h: Hist1d):
+        # Only efficiency is applicable for flat source
```
kdund, line 250: Disagree-- you could have an efficiency defined in true energy and then a smearing on top, I think?

alea/ces_transformation.py
```diff
@@ -0,0 +1,155 @@
...
+        )
+        hist_smeared = Hist1d(data=np.zeros_like(data), bins=hist.bin_edges)
+        hist_smeared.histogram = data
+    else:
+        # use the bins that set by the user
+        bins = np.array(bins)
+        if bins.size <= 1:
+            raise ValueError("bins must have at least 2 elements")
+        bin_centers = 0.5 * (bins[1:] + bins[:-1])
+        data = stats.norm.pdf(
```
kdund, line 44: With coarse binning, I would propose instead defining the data as the difference of the CDF evaluated at the upper edge of the bin and evaluated at the lower edge of the bin

kdund (changes requested):
Hi, @yuema137 awesome to see this PR where we get a dedicated 1D source! Some comments added, I agree with Robert that we should remove specific CES-things to make this useful for any 1D analysis we make :)

alea/ces_source.py
```diff
@@ -0,0 +1,234 @@
...
+        efficiency_transformation = self._create_transformation("efficiency")
+        smearing_transformation = self._create_transformation("smearing")
+        bias_transformation = self._create_transformation("bias")
+
+        # Apply the transformations to the histogram
+        if efficiency_transformation is not None:
+            h = efficiency_transformation.apply_transformation(h)
+        if smearing_transformation is not None:
+            h = smearing_transformation.apply_transformation(h)
+        if bias_transformation is not None:
```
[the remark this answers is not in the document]
yuema137: Actually I think it's better to add in the smeared energy. Because the bias should in principle be caused by some area-dependent effect (like merging SEs and lone hits). For example if a monoenergetic source produces two S2s, 1.1e4 and 1.2 4 separately, the biases would be slightly different for them even if the true energies are the same.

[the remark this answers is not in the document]
kdund: In general the bias should be a function of true parameters (I agree if we had a measurement of "true quanta released" we could define our bias in terms of that, and the CES could be considered an approximation of that, but it is only an approximation)

alea/ces_source.py
kdund: IMO the efficiency should be the probability to detect an event given a certain initial recoil energy. This is what we use for WIMPs, and it is a well-defined probability (that does include the effects you list here as important inputs!). 
I think attempting to apply efficiencies based on reconstructed variables on the basis that those reconstructed variables approximate (but does not give) the underlying number of quanta that are important for the probability to be detected

yuema137: Hi Knut, to clarify, the efficiency here is only projected in CES space. For S2 threshold and S1 efficiency, we get the CES projected values during the band fitting. For cut acceptance which is data-driven, the values are naturally only in CES. Defining everything in true energy is possible but the gain is limited. It will make every efficiency study entangled with the band fitting which is not necessary for lower or other studies performed in CES.

SphericalCowww:
Hello hello Alea experts, let's me join forces to review this PR. 
I am thinking about providing a wiki page recording some tests in the next couple of weeks.

alea/models/blueice_extended_model.py
```diff
@@ -535,7 +548,15 @@ def _set_default_ptype(self):
...
         """
-        allowed_ptypes = ["rate", "shape", "index", "efficiency", "livetime", "needs_reinit"]
+        allowed_ptypes = [
+            "rate",
+            "shape",
+            "index",
+            "efficiency",
+            "livetime",
+            "needs_reinit",
+            "mass",
```
hammannr, line 558: This has no meaning though (i.e. particular feature), right?

alea/ces_transformation.py
```diff
@@ -0,0 +1,157 @@
...
+    smearing_a: float,
+    smearing_b: float,
+    bins: Optional[np.ndarray] = None,
+):
+    """Smear a histogram with Gaussian. This allows for non-uniform histogram binning.
+
+    :param hist: the spectrum we want to smear :param bins: bin edges of the returned spectrum
+    :return: smeared histogram in the same unit as input spectrum
+
+    """
```
hammannr, line 67: Just for consistency in the code, please use the google-style docstring format everywhere.

hammannr (commented):
I had a brief first look at the PR -- very impressive work @yuema137 (but also quite difficult to review because of the many changes). I plan to give it a more thotough review soon and also test a few features. However, I already wanted to share my comments so far.

alea/ces_source.py
```diff
@@ -0,0 +1,490 @@
...
+MINIMAL_ENERGY_RESOLUTION = 0.05
+
+
+def safe_lookup(hist_obj):
+    "A Wrapper to make sure the lookup function returns 0 for out-of-range values"
+    original_lookup = hist_obj.lookup
+
+    def new_lookup(self, *args):
+        if isinstance(self, Hist1d):
+            coordinates = np.asarray(args[0])
```
SphericalCowww, line 22: Also name it "coords" for consistency

alea/ces_source.py
```diff
@@ -0,0 +1,490 @@
...
+            if not in_range.any():
+                return np.zeros_like(coordinates)
+
+            clipped_coords = np.clip(coordinates, self.bin_edges[0], self.bin_edges[-1])
+            result = original_lookup(clipped_coords)
+            result[~in_range] = 0
+
+            return result
+        else:
+            for i, coords in enumerate(args):
```
SphericalCowww, line 34: I would try to avoid single letter variables even in a loop, i.e. maybe "coodIdx" instead of "i". 
This "i" preserves its value deep down in the code and becomes very difficult to pick out.

alea/ces_source.py
```diff
@@ -0,0 +1,490 @@
...
+                coords = np.asarray(coords)
```
SphericalCowww, line 35: Try not to change loop variable values in the middle of a loop, as sometimes it creates an unintended recursion (though more often in a while loop).

alea/ces_source.py
```diff
@@ -0,0 +1,490 @@
...
+
+    Returns:
+    --------
+    Hist1d
+        New histogram with desired binning, properly normalized
+
+    """
+    from scipy.interpolate import interp1d
+    import numpy as np
+    from copy import deepcopy
```
SphericalCowww, line 62: I think these were already imported

alea/ces_source.py
```diff
@@ -0,0 +1,490 @@
...
+
+    # First convert histogram values to density
+    density = hist.histogram / hist.bin_volumes()
+
+    # Create interpolation function for the density
+    f = interp1d(
```
SphericalCowww, line 68: Try to avoid single letter variables, makes ctrl-f much more difficult. Especially "f" is going to be confusing when you uses a lot of f-strings.

alea/ces_source.py
```diff
@@ -0,0 +1,490 @@
...
+        self.templatename = self.config["template_filename"]
+        self.histname = self.config["histname"]
+
+        # Initialize fractions
+        self.fraction_in_range = 1.0
+        self.fraction_efficiency_loss = 1.0
+
+    def _load_true_histogram(self):
+        """Load the true spectrum from the template (no transformation applied)"""
+        h = template_to_multihist(self.templatename, self.histname, hist_to_read=Hist1d)
```
SphericalCowww, line 131: Try to avoid single letter variables, makes ctrl-f much more difficult.

alea/ces_source.py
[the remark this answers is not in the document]
SphericalCowww: I think this would worth a discussion

alea/ces_source.py
```diff
@@ -0,0 +1,490 @@
...
+            h_roi[outside_index_right[0]] *= 1 - frac_right
+
+        if len(outside_index_left) > 0 and self.min_e != 0:
+            frac_left = (self.min_e - left_edges[outside_index_left[-1]]) / (
+                right_edges[outside_index_left[-1]] - left_edges[outside_index_left[-1]]
+            )
+            h_roi[outside_index_left[-1]] *= 1 - frac_left
+
+        h_roi[outside_index_left[:-1]] = 0
+        h_roi[outside_index_right[1:]] = 0
```
SphericalCowww, line 228: I think it's best to have an independent function on this. 
Also preferably using more easily understandable method to derive the result

alea/ces_source.py
```diff
@@ -0,0 +1,490 @@
...
+        inter_pre = interp1d(
+            h_pre.bin_centers, h_pre.histogram, bounds_error=False, fill_value="extrapolate"
+        )
+        inter_bins = np.linspace(
+            h_pre.bin_edges[0],
+            h_pre.bin_edges[-1],
+            int((h_pre.bin_edges[-1] - h_pre.bin_edges[0]) / MINIMAL_ENERGY_RESOLUTION),
+        )
+        h = Hist1d(bins=inter_bins)
+        h.histogram = np.where(inter_pre(h.bin_centers) < 0, 0, inter_pre(h.bin_centers))
```
SphericalCowww, line 265: This is already a function, no?

alea/ces_source.py
```diff
@@ -0,0 +1,490 @@
...
+            frac_right = (right_edges[outside_index_right[0]] - self.max_e) / (
+                right_edges[outside_index_right[0]] - left_edges[outside_index_right[0]]
+            )
+            h.histogram[outside_index_right[0]] *= 1 - frac_right
+
+        if len(outside_index_left) > 0 and self.min_e != 0:
+            frac_left = (self.min_e - left_edges[outside_index_left[-1]]) / (
+                right_edges[outside_index_left[-1]] - left_edges[outside_index_left[-1]]
+            )
+            h.histogram[outside_index_left[-1]] *= 1 - frac_left
```
SphericalCowww, line 283: Again, would be nice in a function

alea/ces_source.py
```diff
@@ -0,0 +1,490 @@
...
+        if method == "linear":
+            if not hasattr(self, "_pdf_interpolator"):
+                # First call:
+                # Construct a linear interpolator between the histogram bins
+                self._pdf_interpolator = interp1d(
+                    self._pdf_histogram.bin_centers,
+                    self._pdf_histogram.histogram,
+                )
+            bcs = self._pdf_histogram.bin_centers
+            clipped_data = np.clip(args, bcs.min(), bcs.max())
```
SphericalCowww, line 358: Applying the rebin function too?

alea/ces_source.py
```diff
@@ -0,0 +1,490 @@
...
+
+            return self._pdf_interpolator(np.transpose(clipped_data))
+
+        elif method == "piecewise":
+            return self._pdf_histogram.lookup(*args)
```
SphericalCowww, line 363: This one is not safe_lookup?

alea/ces_source.py
```diff
@@ -0,0 +1,490 @@
...
+        if self.min_e <= self.mu <= self.max_e:
+            self.fraction_in_range = 1.0
+        else:
+            self.fraction_in_range = 0.0
+
+        # Normalize the histogram
+        h.histogram = h.histogram.astype(np.float64)
+        total_integration = np.sum(h.histogram * h.bin_volumes())
+        if total_integration > 0:
+            h.histogram /= total_integration
```
SphericalCowww, line 449: Perhap also a function just for normalize. Probably can use this function in the function of rebin and normalize too

SphericalCowww (commented):
Hello, I am noob to commenting codes, but here are my suggestions thus far.

alea/ces_transformation.py
```diff
@@ -0,0 +1,157 @@
...
+            * bin_volumes
+        )
+        smeared[idx] = np.sum(probs * rates)
+
+    hist_smeared = Hist1d.from_histogram(smeared, bins)
+
+    return hist_smeared
+
+
+def biasing_hist_arctan(hist: Any, A: float = 0.01977, k: float = 0.01707):
```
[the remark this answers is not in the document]
hammannr: I think we should ask users to always provide those values so I would even suggest getting rid of defaults here alltogether.

alea/ces_transformation.py
```diff
@@ -0,0 +1,157 @@
+from pydantic import BaseModel, validator
+from typing import Dict, Optional, Any, Literal, Callable
+import numpy as np
+from scipy import stats
+from copy import deepcopy
+from multihist import Hist1d
+
+
+def energy_res(energy, a=25.8, b=1.429):
```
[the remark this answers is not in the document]
hammannr: same as above -- probably best to not have defaults that are analysis-specific.

alea/ces_transformation.py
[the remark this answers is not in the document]
SphericalCowww: Also, I was just using the values, I think having a single alphabet "A" and "k" is a bit confusing. Can make them more consistent with the smearing function like "bias_A" and "bias_k", would be the best.

alea/examples/configs/unbinned_ces_simple.yaml
```diff
@@ -0,0 +1,133 @@
...
+      - null
+    parameter_interval_bounds:
+      - 0
+      - 50
+    fit_guess: 1000
+
+  test_flat_rate_multiplier:
+    nominal_value: 1000
+    ptype: rate
+    fittable: ture
```
SphericalCowww, line 29: Hmm..., so "fittable: ture" wouldn't quite get caught by compiler?

alea/examples/configs/unbinned_ces_simple.yaml
```diff
@@ -0,0 +1,133 @@
...
+      - 1.2
+      - 1.6
+    description: smearing shaping parameter
+
+  efficiency_constant:
+    nominal_value: 0.8
+    ptype: shape
+    uncertainty: 0.05
+    relative_uncertainty: false
+    fittable: ture
```
SphericalCowww, line 85: Similarly "ture"

alea/utils.py
```diff
@@ -158,22 +158,26 @@ def get_analysis_space(analysis_space: list) -> list:
...
-    """Prefix file path with template_folder_list whenever possible.
-
-    Args:
-        config (dict): dictionary contains file path
-        template_folder_list (list): list of possible base folders. Ordered by priority.
-        ignore_keys (list, optional (default=["name", "histname"])):
-        keys to be ignored when prefixing
-
-    """
+    """Prefix file path with template_folder_list whenever possible."""
```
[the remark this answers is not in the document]
yuema137: It was accidental. Let me add it back

alea/examples/configs/binned_ces.yaml
```diff
@@ -0,0 +1,143 @@
+parameter_definition:
+  livetime:
+    nominal_value: 365.
+    ptype: livetime
+    fittable: false
+    description: Livetime in day
+
+  xe133_rate_multiplier:
+    nominal_value: 1000
```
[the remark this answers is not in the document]
yuema137: Here, the nominal_value explicitly means the rate in event/ton/year, which I think is more convenient than using a hard-coded nominal rate + nominal value of 1. But for sure we can discuss this point

alea/models/blueice_extended_model.py
```diff
@@ -268,6 +268,11 @@ def _process_blueice_config(self, config, template_folder_list):
         pdf_base_config["livetime_days"] = self.parameters[
             pdf_base_config["livetime_parameter"]
         ].nominal_value
+        pdf_base_config["fiducial_mass"] = (
+            self.parameters[pdf_base_config["fiducial_mass_parameter"]].nominal_value
+            if "fiducial_mass_parameter" in pdf_base_config
+            else 1
+        )
```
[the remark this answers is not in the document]
yuema137: The impact of this parameter is limited to the source-building for the ces sources. It's not passed to any other part in alea or blueice

alea/models/blueice_extended_model.py
yuema137: what exactly are you referring to? It's actually just a reformatting for the code 
for the "mass" term, let me remove

alea/ces_transformation.py
```diff
@@ -0,0 +1,157 @@
...
+
+    :param energy: true energy in keV :return: energy resolution in keV
+
+    """
+    # Reference for the values of a,b:
+    # xenon:xenonnt:analysis:ntsciencerun0:g1g2_update#standard_gaussian_vs_skew-gaussian_yue
+    return (np.sqrt(energy) * a + energy * b) / 100
+
+
+def smearing_mono_gaussian(
```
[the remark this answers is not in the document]
yuema137: we also have `smearing_mono_skew_gaussian` in the LowERAnalysis repo. Here we just show a basic example
For the multiple gaussian smearing, it's also in the LowerAnalysis repo

alea/ces_transformation.py
```diff
@@ -0,0 +1,157 @@
...
+            scale=energy_res(peak_energy, smearing_a, smearing_b),
+        )
+        # create an empty histogram with the user-defined binning
+        hist_smeared = Hist1d(data=np.zeros_like(data), bins=bins)
+        hist_smeared.histogram = data
+
+    return hist_smeared
+
+
+def smearing_hist_gaussian(
```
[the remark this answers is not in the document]
yuema137: Same here. It literally means `smearing_hist_gaussian`
because in the LowER repo we also have `smearing_hist_skew_gaussian`

alea/ces_transformation.py
yuema137: Sure will do

alea/examples/configs/unbinned_ces_simple.yaml
yuema137: This is a typo. Let me fix it. But in Python, it will be interpreted as TRUE as it's a string.

alea/examples/configs/unbinned_ces_simple.yaml
yuema137: Same here. It's a typo

alea/examples/configs/binned_ces.yaml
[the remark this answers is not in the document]
hammannr: Both works in principle but it gets very messy quickly when both the template normalization and the rate parameter are non-one. We learned this the hard way with binference and now try to keep these two separate i.e. rates in the templates and relative scaling in the multiplier.
Of course, technically, both will work.

alea/models/blueice_extended_model.py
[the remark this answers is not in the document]
hammannr: Ok perfect. Is it needed though? To me this looks redundant.

alea/models/blueice_extended_model.py
hammannr: Alright then. Usually we use these type to mark parameters with special features and mass is not one of them, that's what i wanted to raise -- so basically suggesting to removing it here, yes.

alea/models/blueice_extended_model.py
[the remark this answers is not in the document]
yuema137: We added that to account for different fiducial masses in SR0 and SR1 (they share the same rate multiplier, but have different scaling)

alea/models/blueice_extended_model.py
yuema137: Removed

alea/ces_transformation.py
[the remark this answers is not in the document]
yuema137: @RobertoBrauno are you suggesting removing all default values in the transformation part? I'm thinking about where to put those values if we do so - any ideas?

alea/ces_source.py
yuema137: Implemented too

alea/ces_source.py
yuema137: Nice suggestion. Implemented.

alea/ces_source.py
yuema137: this is technially not same as the previous normalization funciton, as in the function we not only normalize the hist, but also apply transformations and calculate event_per_day. Given the straightforward calculation here I would keep the current code.

alea/ces_transformation.py
```diff
@@ -0,0 +1,157 @@
...
+    e_true_s = e_true_s[mask]
+    rates = rates[mask]
+    bin_volumes = bin_volumes[mask]
+
+    e_smeared_s = 0.5 * (bins[1:] + bins[:-1])
+    smeared = np.zeros_like(e_smeared_s)
+
+    for idx, e_smeared in enumerate(e_smeared_s):
+        probs = (
+            stats.norm.pdf(
```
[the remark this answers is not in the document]
yuema137: I didn't fully get it- could you elaborate?

alea/ces_source.py
yuema137: We don't need rebining here, as it's interpolation - the interpolated pdf will be used to calculate the unbinned likelihood, so the binning doesn't matter here.
The rebining is only used for binned likelihood, for which we need consistent binning for all sources.

alea/ces_source.py
yuema137: Same as above - the interpolated pdf will be used. No common binning is needed here.

alea/ces_transformation.py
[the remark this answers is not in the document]
SphericalCowww: In my opinion, it's best to put them as the nominal value in the example configuration yaml files.

alea/examples/configs/unbinned_ces_simple.yaml
SphericalCowww, line 1: Exactly here for the default bias_A, bias_k values. Don't need to turn the fitting for the bias on, but when I was working on the example, I thought it would be nice that I know how which parameters to define.

alea/ces_transformation.py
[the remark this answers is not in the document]
SphericalCowww: I am thinking it would be a similar situation for coarse binning to https://github.com/XENONnT/alea/pull/161/files#r1588456835

SphericalCowww (commented):
I am very delayed, but checking out all the past comments.

alea/examples/configs/unbinned_ces_simple.yaml
yuema137: OK, I think it doesn't harm to remove the default value here. Anyway the nominal value and limits need to be read from the yaml config

alea/examples/configs/binned_ces.yaml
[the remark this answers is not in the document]
hammannr: Fair enough, in this case I would suggest using the already implemented `histogram_scale_factor` and use a nominal value for the rate parameter of 1 after all. This is how we did it basically in the past in binference and as long as you don't mix too much and put scaling in both it's still tidy 😊

alea/ces_transformation.py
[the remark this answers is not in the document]
yuema137: OK removed

alea/examples/configs/binned_ces.yaml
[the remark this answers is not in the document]
yuema137: As I have already overridden the whole `build_pdf_histogram` function, I don't really think it's an "already implemented" method for the 1D case. As the config generator we developed for LowERAnalysis depends on this logic, maybe we can just keep it as it is?

alea/ces_transformation.py
```diff
@@ -0,0 +1,157 @@
...
+    """
+    assert isinstance(hist, Hist1d), "Only Hist1d object is supported"
+    true_energy = hist.bin_centers
+    h_bias = deepcopy(hist)
+    bias_derivative = A * k / (1 + k**2 * true_energy**2)
+    h_bias.histogram *= 1 / (1 + bias_derivative)
+    return h_bias
+
+
+def efficiency_hist_constant(hist: Any, efficiency_constant: float):
```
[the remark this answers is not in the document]
yuema137: The situation is that yes, we can vary only a single fit parameter like (eff_mean + efficiency_constant*eff_sigma).
Otherwise, we cannot afford the time of the fitting

alea/ces_source.py
```diff
@@ -0,0 +1,234 @@
...
+from scipy.interpolate import interp1d
+
+from inference_interface import template_to_multihist
+from blueice import HistogramPdfSource, Source
+from blueice.exceptions import PDFNotComputedException
+
+from multihist import Hist1d
+from alea.ces_transformation import Transformation
+
+MINIMAL_ENERGY_RESOLUTION = 0.05
```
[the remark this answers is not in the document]
yuema137: Good suggestion. Already implemented

alea/ces_transformation.py
```diff
@@ -0,0 +1,157 @@
...
+    },
+    "bias": {"arctan": biasing_hist_arctan},
+    "efficiency": {
+        "constant": efficiency_hist_constant,
+    },
+}
+
+
+# input: model name, parameters, transformation mode
+class Transformation(BaseModel):
```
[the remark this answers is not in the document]
yuema137: By design, it's always applied to hist1d

alea/ces_source.py
```diff
@@ -0,0 +1,490 @@
...
+
+
+def safe_lookup(hist_obj):
+    "A Wrapper to make sure the lookup function returns 0 for out-of-range values"
+    original_lookup = hist_obj.lookup
+
+    def new_lookup(self, *args):
```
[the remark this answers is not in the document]
yuema137: Nice catch. Fixed.

alea/ces_source.py
yuema137: fixed.

alea/ces_source.py
yuema137: fixed

alea/ces_source.py
yuema137: fixed

alea/ces_source.py
yuema137: fixed

alea/ces_source.py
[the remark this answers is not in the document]
yuema137: Good suggestion. Implemented. Now it could be set in the config

SphericalCowww:
Hello @yuema137, I was testing with the case of `apply_efficiency: false` and has the following compilation error:
`AttributeError: 'CESMonoenergySource' object has no attribute 'fraction_efficiency_loss'`
Could you give a default value to it? Thanks!

yuema137:
@SphericalCowww For the bias thing, I've done something a little bit better - adding the fit for bias parameter, but only for the constant $B$.  If we let all parameters fitable and the number of anchors is too much, the fitting gets too slow. I believe fitting the constant $B$ only is enough to demonstrate the functionality of the bias transformation and the fitting.

SphericalCowww:
> @SphericalCowww For the bias thing, I've done something a little bit better - adding the fit for bias parameter, but only for the constant B . If we let all parameters fitable and the number of anchors is too much, the fitting gets too slow. I believe fitting the constant B only is enough to demonstrate the functionality of the bias transformation and the fitting.

Absolutely, when I did my test, I only fitted A. I think as a demonstration, it's more than enough.

SphericalCowww (approved):
Good

hammannr (approved):
I briefly went through the current version and I think it looks good. Thanks a lot @yuema137 for your excellent work and @SphericalCowww for your thorough review!! 🚀
We can ignore the two complaints from CodeFactor -- please go ahead with merging! 😊

GiovanniVolta:
Now that we have reached a consensus of checks and test, we can proceed with he merging

PR #162, opened 2024-05-02
Pegasus-based OSG submitter 

FaroutYLq:
Still in debugging after a major refactoring. Features:
- Highly modularized 
- Auto resubmission when failures happen, with maximum trials configurable.
- Fully pegasus based, so when running failed you can just do `pegasus-run` the same workflow.
- Same submission logic, while you will be using `--htcondor` and a bit more fields to specify in running configuration.

### DIFF
--- a/alea/submitters/htcondor.py
+++ b/alea/submitters/htcondor.py
@@ -1,7 +1,3 @@
-# hyperparameters: configuration
-# inputs: templates in the specified local directory
-# outputs: all *.h5 files in the output folder
-
 import subprocess
 import os
 import threading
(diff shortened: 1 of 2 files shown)


alea/submitters/htcondor.py
```diff
@@ -1,12 +1,619 @@
...
 
 class SubmitterHTCondor(Submitter):
-    """Submitter for htcondor cluster.
+    """Submitter for htcondor cluster."""
+
+    def __init__(self, *args, **kwargs):
+        # General start
+        super().__init__(*args, **kwargs)
+        self.name = self.__class__.__name__
+        self.htcondor_configurations = kwargs.get("slurm_configurations", {})
```
hammannr, line 33: Naive question but why do you get the `slurm_configurations` here? I think I'd have a dedicated `htcondor_configurations` in the runner yaml. This could also have additional fields if you need any!

FaroutYLq:
Realized that I pushed the wrong branch, which is a draft one weeks ago. See next PR for the right one.

PR #163, opened 2024-05-03
Refactored Pegasus-based OSG submitter

FaroutYLq:
Corrected the wrong PR #162. 

After a major refactoring. Read for use now. Features:
- Highly modularized compared to `outsource`
- Auto resubmission when failures happen, with maximum trials configurable.
- Fully pegasus based, so when running failed you can just do `pegasus-run` the same workflow and it will only submit the failed jobs.
- Same submission logic, while you will be using `--htcondor` and a bit more fields to specify in running configuration.

Autotest for this submitter is tricky... we don't have a localUniverse in the resources available yet, and I feel like it is impossible to build a Github workflow submitting to OSG. Will skip auto test imho


FaroutYLq:
OK now jobs can finish running in minimal test for a small cevns test.
```
            "program": "${workspaceFolder}/bin/alea-submission",
            "args": [
                "${workspaceFolder}/lqtests/lq_b8_cevns_running.yaml",
                "--computation",
                "discovery_power",
                "--htcondor",
                "--debug"
            ],
```

FaroutYLq:
Next thing: add summary job to collect outputs and make output tarballs, one for results and one for toys. This will be essential when we scale up to >thousands of jobs.

FaroutYLq:
For the records Up to commit d9f9bc7, this is the yaml files I used.
`lq_b8_cevns_statistical_model.yaml`
```
parameter_definition:
  livetime_sr0:
    nominal_value: 1.221  # in ty, ((97.09 + 14.31) / 365)y * 4t, xenon:xenonnt:analysis:available_data_for_sr0
    ptype: livetime
    fittable: false
    description: SR0 livetime in ty

  livetime_sr1:
    nominal_value: 2.337  # in ty, ((189.55 + 1.03 + 6.68) / 365)y * 4t + 16 extra days = 0.18 txyr, xenon:xenon1t:analysis:september_11th
    ptype: livetime
    fittable: false
    description: SR1 livetime in ty

  b8_rate_multiplier:
    nominal_value: 1.0
    ptype: rate
    uncertainty: null
    relative_uncertainty: true
    fit_guess: 1.0
    fittable: true
    fit_limits:
      - 0
      - 100
    parameter_interval_bounds:
      - 0
      - 100

  ac_sr0_rate_multiplier:
    nominal_value: 1.0
    ptype: rate
    uncertainty: 0.05
    relative_uncertainty: true
    fit_guess: 1.0
    fittable: true
    fit_limits:
      - 0
      - 100
    parameter_interval_bounds:
      - 0
      - 100

  ac_sr1_rate_multiplier:
    nominal_value: 1.0
    ptype: rate
    uncertainty: 0.05
    relative_uncertainty: true
    fit_guess: 1.0
    fittable: true
    fit_limits:
      - 0
      - 100
    parameter_interval_bounds:
      - 0
      - 100

  er_rate_multiplier:
    nominal_value: 1.0
    ptype: rate
    uncertainty: 0.21
    relative_uncertainty: true
    fit_guess: 1.0
    fittable: true
    fit_limits:
      - 0
      - 100
    parameter_interval_bounds:
      - 0
      - 100

  nr_rate_multiplier:
    nominal_value: 1.0
    ptype: rate
    uncertainty: 0.58
    relative_uncertainty: true
    fit_guess: 1.0
    fittable: true
    fit_limits:
      - 0
      - 100
    parameter_interval_bounds:
      - 0
      - 100

  t_ly:
    nominal_value: 0
    ptype: shape
    uncertainty: 1
    relative_uncertainty: false
    fittable: true
    blueice_anchors:
      - -3
      - -2
      - -1
      - 0
      - 1
      - 2
      - 3
    fit_limits:
      - -3
      - 3
    description: Light yield morpher

  t_qy:
    nominal_value: 0
    ptype: shape
    uncertainty: 1
    relative_uncertainty: false
    fittable: true
    blueice_anchors:
      - -3
      - -2
      - -1
      - 0
      - 1
      - 2
      - 3
    fit_limits:
      - -3
      - 3
    description: Charge yield morpher

likelihood_config:
  template_folder: null
  likelihood_terms:
    - name: SR0
      default_source_class: alea.template_source.TemplateSource
      likelihood_type: blueice.likelihood.BinnedLogLikelihood
      analysis_space:
        - "cs2": "70.0000000000000000 133.7313537597656250 207.2897796630859375 500.0000000000000000"
        - "s2_shadow_s2_time_shadow": "np.linspace(0, 1, 4)"
        - "s1_bdt_score": "0.0000000000000000 0.2805856252372395 0.4794203769012032 1.0000000000000000"
        - "s2_bdt_score": "0.7844968000000000 0.8689031088980361 0.9082776496669921 1.0000000000000000"
      in_events_per_bin: true
      livetime_parameter: livetime_sr0
      slice_args: {}
      sources:
      - name: b8
        histname: template
        parameters:
          - b8_rate_multiplier
          - t_ly
          - t_qy
        named_parameters:
          - t_ly
          - t_qy
        template_filename: /home/yuanlq/alea_inputs/nt_cevns_templates/v7/template_XENONnT_sr0_signal_cevns_tly_{t_ly:.1f}_tqy_{t_qy:.1f}.h5

      - name: ac_sr0
        histname: template
        parameters:
          - ac_sr0_rate_multiplier
        template_filename: /home/yuanlq/alea_inputs/nt_cevns_templates/v7/template_XENONnT_sr0_ac_cevns.h5

      - name: er
        histname: template
        parameters:
          - er_rate_multiplier
        template_filename: /home/yuanlq/alea_inputs/nt_cevns_templates/v7/template_XENONnT_sr0_er_cevns.h5

      - name: nr
        histname: template
        parameters:
          - nr_rate_multiplier
          - t_ly
          - t_qy
        named_parameters:
          - t_ly
          - t_qy
        template_filename: /home/yuanlq/alea_inputs/nt_cevns_templates/v7/template_XENONnT_sr0_nr_cevns_tly_{t_ly:.1f}_tqy_{t_qy:.1f}.h5

    - name: SR1
      default_source_class: alea.template_source.TemplateSource
      likelihood_type: blueice.likelihood.BinnedLogLikelihood
      analysis_space:
        - "cs2": "70.0000000000000000 119.4675521850585938 157.4414672851562500 500.0000000000000000"
        - "s2_shadow_s2_time_shadow": "np.linspace(0, 1, 4)"
        - "s1_bdt_score": "0.0000000000000000 0.2595951417391419 0.4594176504038993 1.0000000000000000"
        - "s2_bdt_score": "0.7962600000000000 0.8707681714579936 0.9076863660481916 1.0000000000000000"
      in_events_per_bin: true
      livetime_parameter: livetime_sr1
      slice_args: {}
      sources:
      - name: b8
        histname: template
        parameters:
          - b8_rate_multiplier
          - t_ly
          - t_qy
        named_parameters:
          - t_ly
          - t_qy
        template_filename: /home/yuanlq/alea_inputs/nt_cevns_templates/v7/template_XENONnT_sr1_signal_cevns_tly_{t_ly:.1f}_tqy_{t_qy:.1f}.h5

      - name: ac_sr1
        histname: template
        parameters:
          - ac_sr1_rate_multiplier
        template_filename: /home/yuanlq/alea_inputs/nt_cevns_templates/v7/template_XENONnT_sr1_ac_cevns.h5

      - name: er
        histname: template
        parameters:
          - er_rate_multiplier
        template_filename: /home/yuanlq/alea_inputs/nt_cevns_templates/v7/template_XENONnT_sr1_er_cevns.h5

      - name: nr
        histname: template
        parameters:
          - nr_rate_multiplier
          - t_ly
          - t_qy
        named_parameters:
          - t_ly
          - t_qy
        template_filename: /home/yuanlq/alea_inputs/nt_cevns_templates/v7/template_XENONnT_sr1_nr_cevns_tly_{t_ly:.1f}_tqy_{t_qy:.1f}.h5
```
And the running config `lq_b8_cevns_running.yaml`
```statistical_model: alea.models.BlueiceExtendedModel
statistical_model_config: "/home/yuanlq/software/alea/lqtests/lq_b8_cevns_statistical_model.yaml"

poi: b8_rate_multiplier

computation_options:
  discovery_power:
    to_zip:
      {
        generate_values: [{b8_rate_multiplier: 1.}, {b8_rate_multiplier: 0.}],
        n_mc: [100, 200],
      }
    to_vary:
      {
        nominal_values: [
          {livetime_sr0: 1.221, livetime_sr1: 0.5},
        ]
      }
    in_common:
      {
        hypotheses: ["free", "zero", "true"],
        output_filename: "toymc_power_cevns_livetime_{livetime_sr0:.2f}_{livetime_sr1:.2f}_b8_rate_{b8_rate_multiplier:.2f}.h5",
        toydata_mode: "generate_and_store",
        toydata_filename: "toyfile_cevns_livetime_{livetime_sr0:.2f}_{livetime_sr1:.2f}_b8_rate_{b8_rate_multiplier:.2f}.h5",
        n_mc: 100,
        n_batch: 2,
      }

local_configurations:
  template_path: "/project2/lgrandi/binference_common/nt_cevns_templates"
slurm_configurations:
  template_path: "/project2/lgrandi/binference_common/nt_cevns_templates"
  hours: 5
  mem_per_cpu: 4000
  container: "xenonnt-2023.11.1.simg"
  partition: "lgrandi"
  qos: "lgrandi"
  exclude_nodes: null
htcondor_configurations:
  template_path: "/home/yuanlq/alea_inputs/nt_cevns_templates/v7"
  template_tarball_filename: "/home/yuanlq/alea_inputs/template_tarballs/nt_cevns_templates_v7.tar.gz"
  request_cpus: 1
  request_memory: 2000
  request_disk: 2000000
  dagman_maxidle: 100000
  dagman_retry: 2
  dagman_maxjobs: 100000
  pegasus_transfer_threads: 4
  singularity_image: "/cvmfs/singularity.opensciencegrid.org/xenonnt/montecarlo:2024.04.1"
  running_configuration_filename: "/home/yuanlq/software/alea/lqtests/lq_b8_cevns_running.yaml"
  wf_id: "lq_b8_cevns_25"


outputfolder: "/ospool/uc-shared/project/xenon/yuanlq/alea_outputs/b8mini"
loglevel: "DEBUG"
```

FaroutYLq:
Tested on [a mini job](https://github.com/XENONnT/alea/pull/163#issuecomment-2099612108), turns out to have great performance. Workflow here `/scratch/yuanlq/workflows/outputs/lq_b8_cevns_27`. Now the workflow looks like this
image

We are having redundant double outputs: one on dCache and one on ap23 scratch. See this
```
(XENONnT_development) yuanlq@ap23:/scratch/yuanlq/workflows/outputs/lq_b8_cevns_27$ ls
lq_b8_cevns_27-combined_output.tar.gz               toymc_power_cevns_livetime_1.22_0.50_b8_rate_0.00_0.h5
toyfile_cevns_livetime_1.22_0.50_b8_rate_0.00_0.h5  toymc_power_cevns_livetime_1.22_0.50_b8_rate_0.00_1.h5
toyfile_cevns_livetime_1.22_0.50_b8_rate_0.00_1.h5  toymc_power_cevns_livetime_1.22_0.50_b8_rate_1.00_0.h5
toyfile_cevns_livetime_1.22_0.50_b8_rate_1.00_0.h5  toymc_power_cevns_livetime_1.22_0.50_b8_rate_1.00_1.h5
toyfile_cevns_livetime_1.22_0.50_b8_rate_1.00_1.h5
(XENONnT_development) yuanlq@ap23:/scratch/yuanlq/workflows/outputs/lq_b8_cevns_27$ gfal-ls davs://xenon-gridftp.grid.uchicago.edu:2880/xenon/scratch/yuanlq/lq_b8_cevns_27/00/00
/cvmfs/xenon.opensciencegrid.org/releases/nT/development/anaconda/envs/XENONnT_development/bin/gfal-ls:4: DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
  __import__('pkg_resources').run_script('gfal2-util==1.8.1', 'gfal-ls')
alea-run_toymc
combine
lq_b8_cevns_27-combined_output.tar.gz
lq_b8_cevns_running.yaml
lq_b8_cevns_statistical_model_modified.yaml
nt_cevns_templates_v7.tar.gz
run_toymc_wrapper
run_toymc_wrapper.sh
toyfile_cevns_livetime_1.22_0.50_b8_rate_0.00_0.h5
toyfile_cevns_livetime_1.22_0.50_b8_rate_0.00_1.h5
toyfile_cevns_livetime_1.22_0.50_b8_rate_1.00_0.h5
toyfile_cevns_livetime_1.22_0.50_b8_rate_1.00_1.h5
toymc_power_cevns_livetime_1.22_0.50_b8_rate_0.00_0.h5
toymc_power_cevns_livetime_1.22_0.50_b8_rate_0.00_1.h5
toymc_power_cevns_livetime_1.22_0.50_b8_rate_1.00_0.h5
toymc_power_cevns_livetime_1.22_0.50_b8_rate_1.00_1.h5
```

FaroutYLq:
We do have the concern that too many output files and jobs may cause trouble. We will save that for future development:
- Don't combine more than x number of run toymc jobs
- Turn off outputs from individual run toymc job

FaroutYLq:
Turned off run toymc stage out, so it looks like this now
image

FaroutYLq:
Next thing to do: stress test!

FaroutYLq:
One thing I am not sure if we should do something better is [here](https://github.com/FaroutYLq/alea/blob/31fd1627e580251a77dd63ed12468880df7bb17b/alea/submitters/htcondor.py#L111-L113). Currently I required that there is no subdirectories in the template path, which is not true for our shared template folder `/ospool/uc-shared/project/xenon/binference_common`. That means, users have to find a way to dig out the `.h5` files themselves and dump into a folder assigned to be `template_path`, which is bad. Now it should be fixed. People can just do something like this in their config
```
template_path: "/ospool/uc-shared/project/xenon/binference_common/binference_common/nt_cevns_templates/v9"
```
For example, this config is working now:
```
htcondor_configurations:
  template_path: "/ospool/uc-shared/project/xenon/binference_common/binference_common/nt_cevns_templates/v7"
  cluster_size: 1
  request_cpus: 1
  request_memory: 2000
  request_disk: 2000000
  dagman_maxidle: 100000
  dagman_retry: 2
  dagman_maxjobs: 100000
  pegasus_transfer_threads: 4
  singularity_image: "/cvmfs/singularity.opensciencegrid.org/xenonnt/montecarlo:2024.04.1"
  wf_id: "lq_b8_cevns_34"
```

FaroutYLq:
Spotted a bug: if not `toydata_mode: "generate_and_store"` we will have problems. Should be very fixable

FaroutYLq:
Currently it only works for `generate_and_store`. Need a bit more engineering to make it work also for `read` and `generate`

hammannr:
> One thing I am not sure if we should do something better is [here](https://github.com/FaroutYLq/alea/blob/31fd1627e580251a77dd63ed12468880df7bb17b/alea/submitters/htcondor.py#L111-L113). Currently I required that there is no subdirectories in the template path, which is not true for our shared template folder `/ospool/uc-shared/project/xenon/binference_common`. That means, users have to find a way to dig out the `.h5` files themselves and dump into a folder assigned to be `template_path`, which is bad. Now it should be fixed. People can just do something like this in their config
> 
> ```
> template_path: "/ospool/uc-shared/project/xenon/binference_common/binference_common/nt_cevns_templates/v9"
> ```
> 
> For example, this config is working now:
> 
> ```
> htcondor_configurations:
>   template_path: "/ospool/uc-shared/project/xenon/binference_common/binference_common/nt_cevns_templates/v7"
>   cluster_size: 1
>   request_cpus: 1
>   request_memory: 2000
>   request_disk: 2000000
>   dagman_maxidle: 100000
>   dagman_retry: 2
>   dagman_maxjobs: 100000
>   pegasus_transfer_threads: 4
>   singularity_image: "/cvmfs/singularity.opensciencegrid.org/xenonnt/montecarlo:2024.04.1"
>   wf_id: "lq_b8_cevns_34"
> ```

I think it's not super accurate to check if `template_path` has no subdirectories. Instead, we could check if the base folder is the same for all templates -- this is probably a bit more difficult. However, this is fulfilled if we check that `template_path` has no subdirs so we can go with that for now imo.

hammannr:
Actually there might be a pretty easy solution: We can just check if any .h5 filenames are the same since this is what would cause trouble.

hammannr:
Hey @FaroutYLq , unfortunately I still don't get it running on my end so I can't yet test it properly. However, I think we can improve the handling of the template_path and tar-balling a bit. The idea is that `template_path` is just the general location where templates are. This can have subdirectories (for versions and/or for different sources). The only important thing for tarballing is that the basenames of templates are unique, otherwise, we might silently override things.

To check this and to be a bit more clever about what to add to the tar ball I added the `input_files` attribute to the BEM. This is not yet perfect since it will include all files that match the pattern, which might not all be needed, but still better than taring the entire folder so I think we can use it for now.

My idea is that instead of checkinf for subdirs in the template_path we can do something like this:
```python
basenames = [os.path.basename(f) for f in self.statistical_model.input_files]
unique_basenames = set(basenames)
if len(unique_basenames) != len(basenames):
    raise ValueError("Input files must have unique basenames")
```

and then just tar `self.statistical_model.input_files`. What do you think?

FaroutYLq:
Want to add compatibility support for relative path for files in configs. Discussion [here](https://xenonnt.slack.com/archives/D05APNT2F2R/p1715697538338609)

FaroutYLq:
I also found a bug in computing sensitivity… currently it doesn’t know the right place to find the threshold file. Debuggable

FaroutYLq:
Want better handling of wf_id. See discussion [here](https://xenonnt.slack.com/archives/D05APNT2F2R/p1715789567900829)

hammannr:
After following your suggestion to parse an absolute path for the running config to the submission command it worked and ran smoothly!! 🎉 The solution with pegasus is really amazing and much cleaner compared to what we had running in binference. I'm really excited to see this progress, thanks already @FaroutYLq !


A few things I noticed during this first test:
 * The `outputfolder` parsed in the runner is ignored, instead, the outputs seem to be stored in the workflows/outputs directory. Would it be possible to change this?
 * The submission job creates the `transformations.yml` and `STATISTICAL_MODEL_modified.yaml` file in the path from where the job is submitted. Can we either delete them after submission is done or store them together with the templates in the zip to reduce cluttering?
 * (already talked to you about this but for completeness): Submitting a job with the same `wf_id` lets the job fail without a sensible error message. Only having the basename in the config file would make it more maintainable. I like your suggestion of `actual_wf_id = base_wf_id + <computation>+<datetime>`!

FaroutYLq:
Thanks @hammannr. All good suggestions. Let me rank a few things to do from my side, in rough priority:
- [ ] Get sensitivity work by fixing the thresholds file path on job nodes
- [ ] Better `wf_id`for uniqueness 
- [ ] Move out `setup_pegasus.sh` to maybe README or everyone's home
- [ ] A better way to use `output_folder`
- [ ] Push all generated files into the generated folder, rather than where you submit
- [ ] Improve [tarring](https://github.com/XENONnT/alea/pull/163#issuecomment-2109771220) 
- [ ] Support relative path

hammannr:
One small but convenient feature could be to allow `n_batch` to be a dict such that we can write something like `n_batch: {“slurm”: 10, “htcondor”: 100}` in case we work on both servers 😊

FaroutYLq:
> Hey @FaroutYLq , unfortunately I still don't get it running on my end so I can't yet test it properly. However, I think we can improve the handling of the template_path and tar-balling a bit. The idea is that `template_path` is just the general location where templates are. This can have subdirectories (for versions and/or for different sources). The only important thing for tarballing is that the basenames of templates are unique, otherwise, we might silently override things.
> 
> To check this and to be a bit more clever about what to add to the tar ball I added the `input_files` attribute to the BEM. This is not yet perfect since it will include all files that match the pattern, which might not all be needed, but still better than taring the entire folder so I think we can use it for now.
> 
> My idea is that instead of checkinf for subdirs in the template_path we can do something like this:
> 
> ```python
> basenames = [os.path.basename(f) for f in self.statistical_model.input_files]
> unique_basenames = set(basenames)
> if len(unique_basenames) != len(basenames):
>     raise ValueError("Input files must have unique basenames")
> ```
> 
> and then just tar `self.statistical_model.input_files`. What do you think?



> Hey @FaroutYLq , unfortunately I still don't get it running on my end so I can't yet test it properly. However, I think we can improve the handling of the template_path and tar-balling a bit. The idea is that `template_path` is just the general location where templates are. This can have subdirectories (for versions and/or for different sources). The only important thing for tarballing is that the basenames of templates are unique, otherwise, we might silently override things.
> 
> To check this and to be a bit more clever about what to add to the tar ball I added the `input_files` attribute to the BEM. This is not yet perfect since it will include all files that match the pattern, which might not all be needed, but still better than taring the entire folder so I think we can use it for now.
> 
> My idea is that instead of checkinf for subdirs in the template_path we can do something like this:
> 
> ```python
> basenames = [os.path.basename(f) for f in self.statistical_model.input_files]
> unique_basenames = set(basenames)
> if len(unique_basenames) != len(basenames):
>     raise ValueError("Input files must have unique basenames")
> ```
> 
> and then just tar `self.statistical_model.input_files`. What do you think?

I am a bit lost here actually. Will we run into trouble if someone runs an old running config?

FaroutYLq:
> One small but convenient feature could be to allow `n_batch` to be a dict such that we can write something like `n_batch: {“slurm”: 10, “htcondor”: 100}` in case we work on both servers 😊



> One small but convenient feature could be to allow `n_batch` to be a dict such that we can write something like `n_batch: {“slurm”: 10, “htcondor”: 100}` in case we work on both servers 😊

This is a good point. I wonder if it makes sense to also make `n_mc` into a dict, since you know for OSG we will put an astronomical number compared to on slurm?

FaroutYLq:
> The outputfolder parsed in the runner is ignored, instead, the outputs seem to be stored in the workflows/outputs directory. Would it be possible to change this?

After second thought I found it difficult actually.

- For the same running config, when slurm it will really put outputs there. So you want to put a RCC directory there. However, when htcondor, you don't have access to the same directory because you are on ap23  
- It is quite nontrivial to let people put a directory as output folder in OSG, because all directories are read by OSG in gfal format, for example `gsidavs://xenon-gridftp.grid.uchicago.edu:2880/xenon/scratch/yuanlq/lq_b8_cevns_debug-discovery_power-202405231815/00/00/toymc_power_cevns_livetime_1.22_0.50_b8_rate_0.00_0.h5`. I don't have a good way to auto translate the linux directory users experience in ap23 login node to such format.

I don't have a good way out, and I agree this is confusing that the `outputfolder` doesn't really have the output... So I add a warning message that this `outputfolder` will be ignored in next commit.

FaroutYLq:
The warning message will be like this: 
```
WARNING:root:The outputfolder in the running configuration /ospool/uc-shared/project/xenon/yuanlq/alea_outputs/b8mini won't be used in this submission.
WARNING:root:Instead, you should find your outputs at /scratch/yuanlq/workflows/outputs/lq_b8_cevns_debug-discovery_power-202405231908
```

hammannr:
> I am a bit lost here actually. Will we run into trouble if someone runs an old running config?

This is my suggestion to replace the subdir check in the htcondor submitter with this one. Of course, not all old configs will then work on osg but I think that's ok. As long as the tunning on slurm is not affected we can do that!

The `model.input_files` are a thing I added for everyone. Right now they're not needed for anything else but you can use them in the htcondor submitter to check if we can easily combine all files into one tar without any of them being overwritten (i.e. if any input files share the same base folder and are only unique due to their subdir.).

hammannr:
> > One small but convenient feature could be to allow `n_batch` to be a dict such that we can write something like `n_batch: {“slurm”: 10, “htcondor”: 100}` in case we work on both servers 😊
> 
> > One small but convenient feature could be to allow `n_batch` to be a dict such that we can write something like `n_batch: {“slurm”: 10, “htcondor”: 100}` in case we work on both servers 😊
> 
> This is a good point. I wonder if it makes sense to also make `n_mc` into a dict, since you know for OSG we will put an astronomical number compared to on slurm?

Yes good idea! 👍

hammannr:
> > The outputfolder parsed in the runner is ignored, instead, the outputs seem to be stored in the workflows/outputs directory. Would it be possible to change this?
> 
> After second thought I found it difficult actually.
> 
> * For the same running config, when slurm it will really put outputs there. So you want to put a RCC directory there. However, when htcondor, you don't have access to the same directory because you are on ap23
> * It is quite nontrivial to let people put a directory as output folder in OSG, because all directories are read by OSG in gfal format, for example `gsidavs://xenon-gridftp.grid.uchicago.edu:2880/xenon/scratch/yuanlq/lq_b8_cevns_debug-discovery_power-202405231815/00/00/toymc_power_cevns_livetime_1.22_0.50_b8_rate_0.00_0.h5`. I don't have a good way to auto translate the linux directory users experience in ap23 login node to such format.
> 
> I don't have a good way out, and I agree this is confusing that the `outputfolder` doesn't really have the output... So I add a warning message that this `outputfolder` will be ignored in next commit.

Ah ok I was not aware -- then your solution with the warning is perfect!

FaroutYLq:
Now we support all relative paths as supported by slurm mode!

FaroutYLq:
For supporting dictionary `n_mc` and `n_batch` I met with some issue. It seems to me in the parent `Submitter` there is no way to know which way we will submit... I don't have a quick and elegant solution for it right now...

FaroutYLq:
@hammannr now I implemented the filename check you suggested, but not still too sure how we use the `model.input_files`... Except for that I think we can aim for some tweaking and then merging

dachengx:
I suggest you make a branch inside alea repo next time submitting PR. @FaroutYLq :)

alea/models/blueice_extended_model.py
```diff
@@ -333,6 +337,10 @@ def _build_ll_from_config(
         # Iterate through each likelihood term in the configuration
         for config in likelihood_config["likelihood_terms"]:
             blueice_config = self._process_blueice_config(config, template_folder_list)
+            for source in blueice_config["sources"]:
```
dachengx, line 340: `blueice_config["sources"]` -> `blueice_config["pdf_base_config"]["sources"]`

dachengx (changes requested):

FaroutYLq:
@hammannr @dachengx I updated the setup file instruction, since the situation in ap23 changed a bit since I worked on this last time, due to the EL9 upgrade. I think now ap23 will be quite stable for years.

alea/models/blueice_extended_model.py
FaroutYLq: @hammannr can you please explain a bit here? Thanks

dachengx:
What is the reason to add `self.input_files` https://github.com/FaroutYLq/alea/commit/dad23314fb48048b505f50be02dfb4e15199571d? @hammannr

FaroutYLq:
> What is the reason to add `self.input_files` [FaroutYLq@dad2331](https://github.com/FaroutYLq/alea/commit/dad23314fb48048b505f50be02dfb4e15199571d)? @hammannr

There was some optimization @hammannr was aiming for, but not sure if completed (my vague memory says no). I feel like at this stage it might be better to tweak over tests and merge first, then implement other optimization and features.

dachengx (approved):
We should merge like this.

FaroutYLq:
Happy to get it merged. In principle we should avoid super PR with too many new features squashed.

PR #165, opened 2024-06-07
Debug for interpolator deduction of `NeymanConstructor`

dachengx:


alea/submitters/local.py
```diff
@@ -472,7 +478,6 @@ def get_confidence_interval_thresholds(
                 )
 
             # get the poi_values and threshold_values for interp1d
-            interpolator = None
```
dachengx, line 475: I do not understand why we have this line. So remove it.

hammannr (approved):
Looks good, thanks @dachengx !

Issue #167, opened 2024-06-17
Try prefix all filename with a pre-defined path


dachengx:
In the `.yaml` configuration file, if a filename key can not be recognized by alea, we must manually prefix a path. Like this `binning_reference_filename` is not recognized automatically so I need to write absolute path of it:

```
binning_reference_filename: /project2/lgrandi/binference_common/binning_reference_filename.h5
```

PR #169, opened 2024-06-20
Try prefix every file path in likelihood configuration with template folder

dachengx:
close: https://github.com/XENONnT/alea/issues/167

1. go through `likelihood_config`, for `str` in its values, try prefix template folder to them
2. go through every source in `likelihood_config_copy["sources"]`, for `str` in its values, try prefix template folder to them

### DIFF
--- a/alea/utils.py
+++ b/alea/utils.py
@@ -117,6 +117,22 @@ def get_analysis_space(analysis_space: list) -> list:
     return eval_analysis_space
 
 
+def _prefix_file_path(config: dict, template_folder_list: list):
+    """Prefix file path with template_folder_list whenever possible.
+
+    Args:
+        config (dict): dictionary contains file path
+        template_folder_list (list): list of possible base folders. Ordered by priority.
+
+    """
+    for key in config.keys():
+        if isinstance(config[key], str):
+            try:
+                config[key] = get_file_path(config[key], template_folder_list)
+            except RuntimeError:
+                pass
+
+
 def adapt_likelihood_config_for_blueice(
     likelihood_config: dict, template_folder_list: list
 ) -> dict:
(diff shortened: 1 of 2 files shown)


dachengx:
Some numerical error can happen: https://github.com/XENONnT/alea/actions/runs/9590369297/job/26445626013?pr=169 bt maybe we can return to it later.

alea/utils.py
```diff
@@ -117,6 +117,22 @@ def get_analysis_space(analysis_space: list) -> list:
...
+        config (dict): dictionary contains file path
+        template_folder_list (list): list of possible base folders. Ordered by priority.
+
+    """
+    for key in config.keys():
+        if isinstance(config[key], str):
+            try:
+                config[key] = get_file_path(config[key], template_folder_list)
+            except RuntimeError:
+                pass
```
hammannr, line 133: When would you get the RuntimeError?

hammannr (commented):
In principle this looks good, we need to be careful that we're able to track which files are being loaded but mostly this is equivalent to what was done before anyways.
@FaroutYLq this PR might screw up some things in #163, though I think fixable.

alea/utils.py
dachengx: Here https://github.com/XENONnT/alea/blob/4a6f813399eca557ab21a9b198454f2e7c08d2bb/alea/utils.py#L283

hammannr: Ok so basically whenever the str is not a file (or it can't find it but then it will raise an error later)? Makes sense then to silence this.

dachengx: Yes, I agree.

hammannr (approved):
Let's merge it then! Thanks, @dachengx !

PR #178, opened 2024-07-23
Document to increase CPUs

FaroutYLq:
This is to fit the new situation on OSG, which forbids excessive CPU usage. Otherwise, typically error will be 
```
Excessive CPU usage. Please verify that the code is configur
ed to use a limited number of cpus/threads, and matches request_cpus.
```

However, I didn't figure out how many threads do `alea-run_toymc` is trying to open. I just put an empirical 4 here for now. Still testing...


hammannr:
@FaroutYLq we can merge this as a first try. Basically this means we should test a bit with 4 and see what happens? Anything particular to look out for?

hammannr (approved):

FaroutYLq:
I already tested, and 4 CPUs generally work. (On the other hand, 1 CPU doesn't work anymore! Be alarmed). It would be better if we can find where multithreading happen in alea, but it might take a while and might be overkill. (I am not sure if OSG gonna keep the CPU rule for very long, so I think it doesn't worth the efforts now). Let's blindly do this for now: )

PR #181, opened 2024-07-29
Conditional Parameters

hammannr:
This PR introduces conditional parameters.

### What's the problem?
In some cases, we want parameters to take different values depending on the value of other parameters.

One example is the WIMP-mass-dependent signal efficiency uncertainty. Here, we typically have larger uncertainties for lower masses. With conditional parameters, we can now simply add the following to our parameter definition in the config file:

```
wimp_mass:
    nominal_value: 50
    ...
signal_efficiency:
    conditioning_parameter_name: wimp_mass
    uncertainty:
      5: 0.3
      10: 0.2
      50: 0.15
      100: 0.1
      500: 0.1
    ...
```

Depending on the value of `wimp_mass` this will result in a different value of the `signal_efficiency` uncertainty.

### The ConditionalParameter class
Here's an example how the ConditionalParameter class works:

```python
from alea import load_yaml, Parameters


config_path = "%SOMEPATH%/alea/alea/examples/configs/unbinned_wimp_statistical_model_mass_dependent_efficiency.yaml"
config = load_yaml(config_path)

params = Parameters.from_config(config["parameter_definition"])

#  If you directly access a property of the conditional parameter, it will just return the
#  Parameter under nominal conditions (i.e. using the uncertainty for wimp_mass = 50)
print(params.signal_efficiency.uncertainty)
# >>> 0.15

#  The same happens if you call the conditional parameter without any arguments:
print(params.signal_efficiency().uncertainty)
# >>> 0.15

#  If you provide another value though, it will return the parameter for another wimp_mass:
print(params.signal_efficiency(wimp_mass=5).uncertainty)
# >>> 0.3

#  Another way to get there is to change the nominal value of wimp_mass:
params.wimp_mass.nominal_value = 5
print(params.signal_efficiency.uncertainty)
# >>> 0.3
```

### Caveat
So far, this doesn't work "on-the-fly" i.e. if I have a model instance and just change the `wimp_mass` value it won't correctly update the treatment of the signal efficiency since the likelihood is already initialized with the constraint terms. However, the structure introduced in this PR should allow introducing this feature in the future.

### What I tested so far
I tested this in a notebook for a few different WIMP masses (this won't work with the config in alea since we don't have any other WIMP mass templates yet, but you can ask me for an example notebook using actual templates). So far, everything looked good and behaved as expected but heavier testing with large toyMCs will follow.

Please have a look, play around with the new class, and let me know if something unexpected happens. Also, if you have any suggestions on how to implement this better, please let me know! In the meantime, I'll write some unittests and do some first checks with toyMCs.

### DIFF
--- a/alea/examples/configs/unbinned_wimp_statistical_model_mass_dependent_efficiency.yaml
+++ b/alea/examples/configs/unbinned_wimp_statistical_model_mass_dependent_efficiency.yaml
@@ -1,0 +1,144 @@
+parameter_definition:
+  wimp_mass:
+    nominal_value: 50
+    fittable: false
+    description: WIMP mass in GeV/c^2
+
+  livetime_sr0:
+    nominal_value: 0.2
+    ptype: livetime
+    fittable: false
+    description: Livetime of SR0 in years
+
+  livetime_sr1:
+    nominal_value: 1.0
+    ptype: livetime
+    fittable: false
+    description: Livetime of SR1 in years
+
+  wimp_rate_multiplier:
+    nominal_value: 1.0
+    ptype: rate
+    fittable: true
+    fit_limits:
+      - 0
+      - null
+    parameter_interval_bounds:
+      - 0
+      - null
+
+  er_rate_multiplier:
+    nominal_value: 1.0
+    ptype: rate
+    uncertainty: 0.2
+    relative_uncertainty: true
+    fittable: true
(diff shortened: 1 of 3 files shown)


FaroutYLq:
Hi thanks @hammannr. How urgent is this one? Is it OK if I come back to it this Friday?

PR #182, opened 2024-07-30
Enable choosing the fit strategy

hammannr:
### What's the issue?
In the past, we observed that sometimes, and under certain conditions, minuit might not find a "valid" minimum for the fit. There are a few different solutions to this issue:
 * Continue the fit with higher precision
 * Use a different minimizer routine to begin with
 * Increase fit precision to begin with

### What does this PR add?
In this PR, we enable choosing a fit strategy. For this, we add `_fit_strategy` attribute to the `StatisticalModel` class, which is used in the `fit()` and `confidence_interval()` method if not specified otherwise.
The `_fit_strategy` is a dictionary with the following entries:

| Entry | Functionality | Options |
|:-----|:---|:---|
|`minimizer_routine` (Default: `"migrad"`) | the minimizer routine to use | "migrad", "simplex", or "simplex_migrad" (first run simplex, then migrad) |
|`minuit_strategy` (Default: `1`) | "strategy" property of Minuit. The higher the number, the more precise the fit but also the slower. | 0, 1, or 2|
|`refit_invalid` (Default: `True`) | if True, refit with the `simplex_migrad` routine and strategy `2` if the optimization does not converge the first time.| True/False|
|`disable_index_fitting` (Default: `False`) |If True, disable the index fitting introduced in #156 even if the model has index parameters. | True/False|
|`max_index_fitting_iter` (Default: `10`) |Maximum number of iterations for index fitting | any integer|

The last two are not new but were direct arguments of the `fit` method before. 

There are a few ways to override the defaults on different levels:
 * Add `fit_strategy` with all/some overrides to your running config file and the strategy will be used for every computation of the runner
 * Initialize the statistical model with custom `fit_strategy` dict.
 * For an individual fit or confidence interval computation, pass a custom strategy to the method call. This can for example be used to compare performance and time requirements for different strategies before committing to one of them.

### How does the default behavior change?
The only difference in the default behavior is that invalid minima are now refit by default, which was not possible before. So far, the Minuit strategy was not specified and the Minuit default of 1 was used, which is the default also now. Also, the standard minimizer routine remains `migrad` as before.

### DIFF
--- a/alea/examples/configs/unbinned_wimp_running.yaml
+++ b/alea/examples/configs/unbinned_wimp_running.yaml
@@ -66,13 +66,3 @@ htcondor_configurations:
   singularity_container: null
 
 outputfolder: null
-
-# just for illusrative purposes, the default values are shown here.
-# they are not needed in the configuration file but you could
-# overwrite them here.
-fit_strategy:
-  disable_index_fitting: False
-  max_index_fitting_iter: 10
-  minimizer_routine: "migrad"
-  minuit_strategy: 1
-  refit_invalid: True
--- a/alea/examples/configs/unbinned_wimp_running.yaml
+++ b/alea/examples/configs/unbinned_wimp_running.yaml
@@ -66,3 +66,13 @@ htcondor_configurations:
   singularity_container: null
 
 outputfolder: null
+
+# just for illusrative purposes, the default values are shown here.
+# they are not needed in the configuration file but you could
+# overwrite them here.
+fit_strategy:
+  disable_index_fitting: False
+  max_index_fitting_iter: 10
+  minimizer_routine: "migrad"
+  minuit_strategy: 1
+  refit_invalid: True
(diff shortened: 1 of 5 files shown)


dachengx (approved):
Looks good. But then you also need to update the OSG submitter.

hammannr:
> Looks good. But then you also need to update the OSG submitter.

@dachengx I only modified the submitter base class since all other submitters will parse additional arguments tot the base class or am I missing something? Unfortunately, I can't test this behaviour though.

hammannr:
@dachengx at least on midway this seems to be working fine. Any reason why it shouldn't work for the osg submitter? Maybe also @FaroutYLq can have a guick look 😊

FaroutYLq:
@hammannr I think it should work for OSG, since in principle the new `fit_strategy` things should just show up in the ticket generator outputs, and collected [here](https://github.com/XENONnT/alea/blob/5c1ee1e98e7a417c50ee2616acda37e80cb5068a/alea/submitters/htcondor.py#L558-L559). I didn't do test though (and it is nontrivial to test on OSG until this alea is in environment)

hammannr:
After some internal discussions, we decided that the standard behavior should be that the strategy is defined in the config file of the statistical model and that it can be overwritten by values in the runner. I added this functionality now i.e. runner fit strategy > stat. model fit strategy > default fit strategy.

PR #189, opened 2024-08-26
Fix OSG submission

hammannr:
In #182  we added `fit_strategy` as an argument to the runner. Unfortunately, I missed to also add it to the shell script, which is called for submissions to OSG, which led to shifted arguments. This PR should fix the mismatch.


FaroutYLq (commented):
Thanks for the headup. It looks correct!

PR #198, opened 2024-09-05
There was a typo in docstr

FaroutYLq:
- for memory, the unit is MB
- for disk, the unit is KB


dachengx:
I think we'd better use MB all time.

FaroutYLq:
Sure I agree. Maybe we wrap up in alea (and also outsource) into MB-only units. We probably cannot do anything for Pegasus.

PR #201, opened 2024-09-05
Use MB all the time in `SubmitterHTCondor`, no more kB

dachengx:
More than https://github.com/XENONnT/alea/pull/198

[slack thread](https://xenonnt.slack.com/archives/C04JRF0AZTP/p1725488333322409) (internal reference)


dachengx:
Because https://github.com/XENONnT/alea/pull/198#issuecomment-2330515466, we are fine to merge.

Issue #204, opened 2024-09-06
`toydata_mode` in `SubmitterHTCondor` does not support `read` or `no_toydata`


dachengx:
As the title described.

https://github.com/XENONnT/alea/blob/832c5d773bbf6d10d999960decc2387860d83683/alea/submitters/htcondor.py#L599-L602

PR #206, opened 2024-09-06
Support more `toydata_mode`

dachengx:
Close: https://github.com/XENONnT/alea/issues/204


FaroutYLq (commented):
I feel like this needs a bit more consideration. For example, when you submit a "read" job, you may have to upload a ton of input files, which is dangerous. The more practical solution will be to tar inputs up first when transferring. (Remember, you will need to also untar them in the .sh scripts which are supposed to be running on job node.)

dachengx:
> I feel like this needs a bit more consideration. For example, when you submit a "read" job, you may have to upload a ton of input files, which is dangerous. The more practical solution will be to tar inputs up first when transferring. (Remember, you will need to also untar them in the .sh scripts which are supposed to be running on job node.)

Usually, we will not assign different `toydata_filename` for various jobs, so the number of `toydata_filename` will not be so large. I think we are fine to keep it like this, and come back if problem happens.

PR #213, opened 2024-09-19
Use utilix to validate X509 proxy

dachengx:
Depends on https://github.com/XENONnT/utilix/pull/124.


dachengx:
Actually, this PR depends on a new release of https://github.com/XENONnT/utilix, but we never tested the `SubmitterHTCondor` in the `alea` repo so there is no failure in the auto-test.

PR #214, opened 2024-09-19
Tarball all needed templates from different folder

dachengx:
as long as their names are different

Asking all templates are in the same folder does not make sense, configuration like: https://github.com/XENONnT/nton/blob/fcccf25d583e1ee218724c4962d6fac9583c5191/nton/wimp/configs/sr0_sr1/wimp_statistical_model_ac_optimization.yaml#L409 and https://github.com/XENONnT/nton/blob/fcccf25d583e1ee218724c4962d6fac9583c5191/nton/wimp/configs/sr0_sr1/wimp_statistical_model_ac_optimization.yaml#L553 should be allowed.

1. Add a class `LockableSet` and initialize a variable `TEMPLATE_RECORDS` in `alea/utils.py`
2. Update `TEMPLATE_RECORDS` in `alea.utils._prefix_file_path`, so `TEMPLATE_RECORDS` will record all needed templates absolute path
3. Then we do not need `alea.submitters.htcondor.SubmitterHTCondor._validate_template_path` or `alea.submitters.htcondor.SubmitterHTCondor._check_filename_unique` any more
4. In `alea.submitters.htcondor.SubmitterHTCondor._tar_h5_files`, first copy all templates into `templates_tarball_dir` then make tarball of `templates_tarball_dir`

### DIFF
--- a/alea/submitters/htcondor.py
+++ b/alea/submitters/htcondor.py
@@ -25,7 +25,7 @@ from Pegasus.api import (
 )
 from alea.runner import Runner
 from alea.submitter import Submitter
-from alea.utils import RECORDS, load_yaml, dump_yaml
+from alea.utils import TEMPLATE_RECORDS, load_yaml, dump_yaml
 
 
 DEFAULT_IMAGE = "/cvmfs/singularity.opensciencegrid.org/xenonnt/base-environment:latest"
@@ -70,7 +70,7 @@ class SubmitterHTCondor(Submitter):
         self.dagman_maxjobs = self.htcondor_configurations.pop("dagman_maxjobs", 100_000)
 
         super().__init__(*args, **kwargs)
-        RECORDS.lock()
+        TEMPLATE_RECORDS.lock()
 
         # Job input configurations
         self.config_file_path = os.path.abspath(self.config_file_path)
@@ -135,7 +135,7 @@ class SubmitterHTCondor(Submitter):
 
     def _make_template_tarball(self):
         """Make tarball of the templates if not exists."""
-        if not RECORDS.uniqueness:
+        if not TEMPLATE_RECORDS.uniqueness:
             raise RuntimeError("All files in the template path must have unique basenames.")
         os.makedirs(self.templates_tarball_dir, exist_ok=True)
         if os.listdir(self.templates_tarball_dir):
@@ -145,7 +145,7 @@ class SubmitterHTCondor(Submitter):
             )
 
         logger.info(f"Copying templates into {self.templates_tarball_dir}")
-        for record in tqdm(RECORDS):
+        for record in tqdm(TEMPLATE_RECORDS):
             # Copy each file to the destination folder
             shutil.copy(record, self.templates_tarball_dir)
         self._tar_h5_files(self.templates_tarball_dir, self.template_tarball)
--- a/alea/submitters/htcondor.py
+++ b/alea/submitters/htcondor.py
@@ -4,7 +4,9 @@ import tarfile
 import shlex
 from datetime import datetime
 import logging
+import shutil
 from pathlib import Path
+from tqdm import tqdm
 from utilix.x509 import _validate_x509_proxy
 from Pegasus.api import (
     Arch,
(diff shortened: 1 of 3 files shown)


dachengx:
Actually this PR depends on a new release of https://github.com/XENONnT/utilix, but we never tested the `SubmitterHTCondor` in `alea` repo so there I no failure is the auto-test.

alea/submitters/htcondor.py
```diff
@@ -123,41 +127,29 @@ def requirements(self):
...
-        for dirpath, dirnames, filenames in os.walk(self.template_path):
-            for filename in filenames:
-                logger.info(f"File: {filename} in {dirpath}")
-        if self._contains_subdirectories(self.template_path):
-            logger.warning(
-                "The template path contains subdirectories. All templates files will be tarred."
-            )
-
     def _tar_h5_files(self, directory, template_tarball="templates.tar.gz"):
         """Tar all .h5 templates in the directory and its subdirectories into a tarball."""
```
hammannr, line 131: The docstring should be updated

alea/submitters/htcondor.py
```diff
@@ -123,41 +127,29 @@ def requirements(self):
...
-            # Walk through the directory
-            for dirpath, dirnames, filenames in os.walk(directory):
-                for filename in filenames:
-                    if filename.endswith(".h5"):
-                        # Get the full path to the file
-                        filepath = os.path.join(dirpath, filename)
-                        # Add the file to the tar
-                        # Specify the arcname to store relative path within the tar
-                        tar.add(filepath, arcname=os.path.basename(filename))
+            tar.add(directory, arcname=os.path.basename(directory))
```
hammannr, line 134: This could in principle add things that are not needed (i.e. everything not ending on .h5, like pdf_cache or something), right? I also have a faint memory of Tim choosing the flat folder structure because otherwise it wouldn't work on the computing node but I might be wronog -- have you checked though?

alea/submitters/htcondor.py
```diff
@@ -715,6 +692,6 @@ def submit(self, **kwargs):
             self.wf.graph(
                 output=os.path.join(self.generated_dir, "workflow_graph.dot"), label="xform-id"
             )
-            self.wf.graph(
-                output=os.path.join(self.generated_dir, "workflow_graph.svg"), label="xform-id"
-            )
+            # self.wf.graph(
+            #     output=os.path.join(self.generated_dir, "workflow_graph.svg"), label="xform-id"
+            # )
```
hammannr, line 697: ?

alea/submitters/htcondor.py
```diff
@@ -123,41 +127,29 @@ def requirements(self):
...
-        # This path must exists locally, and it will be used to stage the input files
-        if not os.path.exists(self.template_path):
-            raise ValueError(f"Path {self.template_path} does not exist.")
-
-        # Printout the template path file structure
-        logger.info("Template path file structure:")
-        for dirpath, dirnames, filenames in os.walk(self.template_path):
-            for filename in filenames:
-                logger.info(f"File: {filename} in {dirpath}")
-        if self._contains_subdirectories(self.template_path):
```
hammannr, line 139: `_contains_subdirectories` is no longer used after this PR sop we might as well delete it or am I missing something?

hammannr (commented):
Really nice solutionwith the lockable set! Looks good to me in general but I haven't tried it (yet). One small concern are the non-flat structures in the tar ball but I might be wrong here.
One question not entirely related to this PR: Shouldn't the toyfiles be included here as well? Since we no allow toydata mode "read" they should be tarred as well or am I missing something?

alea/submitters/htcondor.py
dachengx: If you run into the code, you will find that only templates are copied to `self.templates_tarball_dir`.

alea/submitters/htcondor.py
dachengx: Because some large submission need too long time for making the `svg` structure of workflow.

alea/submitters/htcondor.py
dachengx: Yes, you are right.

dachengx:
> Really nice solutionwith the lockable set! Looks good to me in general but I haven't tried it (yet). One small concern are the non-flat structures in the tar ball but I might be wrong here. One question not entirely related to this PR: Shouldn't the toyfiles be included here as well? Since we no allow toydata mode "read" they should be tarred as well or am I missing something?

Toydata is add to the file in this line: https://github.com/XENONnT/alea/blob/ba824ac2c2ffb5e5e843d619b7f5fd9d31141d05/alea/submitters/htcondor.py#L593.

Issue #221, opened 2024-10-01
The `self.scratch_dir` is invalid `Directory.SHARED_SCRATCH`


dachengx:
Because later we set `f"gsidavs://xenon-gridftp.grid.uchicago.edu:2880/xenon/scratch/{getpass.getuser()}"` as `Directory.SHARED_SCRATCH`: https://github.com/XENONnT/alea/blob/356c4b18b991e0386644a4816202df431859ab4d/alea/submitters/htcondor.py#L249, the `self.scratch_dir` in https://github.com/XENONnT/alea/blob/356c4b18b991e0386644a4816202df431859ab4d/alea/submitters/htcondor.py#L216 is never used.

FaroutYLq:
You are right. Please feel free to remove `scratch_dir`.

PR #222, opened 2024-10-01
Keep only one `Directory.SHARED_SCRATCH`

dachengx:
Close: https://github.com/XENONnT/alea/issues/221

Remove `self.scratch_dir` because it is not needed. Generally speaking, for pegasus workflow building we only need a single `Directory.SHARED_SCRATCH` and a single `Directory.LOCAL_STORAGE`.


FaroutYLq (approved):
It's clear enough. Thanks.

PR #225, opened 2024-10-04
Add file hash to blueice hash

hammannr:
So far, if you made changes to a template that was already cached by blueice, blueice would use the cached version instead of the new one. This PR solves this issue by computing the hash of each source's template file and adding it to the blueice source config.

### DIFF
--- a/alea/template_source.py
+++ b/alea/template_source.py
@@ -7,7 +7,7 @@ from blueice import HistogramPdfSource
 from multihist import Hist1d
 from inference_interface import template_to_multihist
 
-from alea.utils import load_json, compute_file_hash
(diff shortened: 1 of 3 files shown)


hammannr:
I checked that it works but quickly want to add a unittest and do a few more checks tomorrow.

hammannr:
But if you already have any code-related suggestions @dachengx please let me know 😊

hammannr:
Ohhh ok nvm -- it kills some unittests, I'll fix it tomorrow..

dachengx:
For the pre-commit problem, you can either merge while ignoring it or implement this: https://github.com/PyCQA/docformatter/pull/287#issuecomment-2401028403

dachengx (approved):

PR #228, opened 2024-10-11
Source-wise interpolation

hammannr:
Enables (and defaults to) source-wise interpolation, which is introduced in [blueice PR46](https://github.com/JelleAalbers/blueice/pull/46).

Let's wait with this PR until the blueice PR is done but I'll put this here in case anyone wants to already give it a try!


kdund (approved):
Test pass, minimal changes-- thank you, @hammannr !

PR #231, opened 2024-10-17
Tar user-installed blueice to osg nodes

hammannr:
Similar to alea, this now tars and installs a user-installed version of blueice to osg computing nodes.


dachengx (approved):

dachengx (dismissed):

alea/submitters/run_toymc_wrapper.sh
```diff
@@ -93,6 +93,7 @@ METADATA=$(echo "$metadata" | sed "s/'/\"/g")
 
 # Installing customized packages
 . install.sh alea
```
dachengx, line 95: This can also be `. install.sh alea blueice`.

hammannr: done!

PR #236, opened 2024-12-11
Update blueice dependence

hammannr:
This updates blueice dependence to require version 1.2.1, which includes the source-wise interpolation.


dachengx (approved):
Thank you very much, @hammannr !

Issue #235, opened 2024-12-11
Generalise SLURM submission


kdund:
Alea currently provides a slurm submitter that uses xenons utilix/batchq. 
However, this uses and assumes the xenon batch environment-- e.g. here: 
https://github.com/XENONnT/utilix/blob/9134ad95de0e9bc29178ae331d2fe241bd45113b/utilix/batchq.py#L43

We should move this submitter to a slurm_xenon submitter and change the "standard" SLURM submitter to not enforce these constraints

mdoerenkamp:
The same assumptions are made for the HTCondor submitter. Please also add a generalised version of this

PR #238, opened 2025-01-22
Pared-down SLURM submitter for non-XENON useage

kdund:
This moves the XENONnT midway slurm submitter to rcc_slurm and makes a new (copy-pasted from the old with bits removed) slurm submitter 
addresses #235


hammannr:
What's the status on this PR @kdund? Are you just waiting for review? Probably this would be a good time to wrap it up 😊

kdund:
> What's the status on this PR @kdund? Are you just waiting for review? Probably this would be a good time to wrap it up 😊

Precisely, i think I had someone lined up but they were busy— if you’re able or have someone this would be super! (And good for presenting to others :) )

hammannr:
I can try to find someone or otherwise I can test it myself 👍

hammannr:
I managed to run this on the MPIK cluster -- very neat! Some annoyances with utilix overhead, but I think it is still better to live with this than make something custom, at least for now. I'll have a more thorough look at the code tomorrow/Thursday, and then we can hopefully merge soon! 😊

alea/submitters/slurm.py
```diff
@@ -1,27 +1,26 @@
 import os
 import time
 import inspect
```
hammannr, line 3: This can be removed now since it's not used here but only in `rcc_slurm.py`

alea/submitters/slurm.py
```diff
@@ -1,27 +1,26 @@
...
+import tempfile
+import datetime
 
+
+from typing import Any, Dict, List, Literal, Optional
```
hammannr, line 8: `Literal` is not used, I guess this is a remnant from the development.

alea/submitters/rcc_slurm.py
```diff
@@ -0,0 +1,109 @@
...
+    "mem_per_cpu": 2000,  # in the unit of Mb
+    "container": "xenonnt-development.simg",
+    "partition": "xenon1t",
+    "qos": "xenon1t",
+    "cpus_per_task": 1,
+    "exclude_nodes": "dali[028-030],midway2-0048",
+}
+
+
+class SubmitterRCCSlurm(Submitter):
```
hammannr, line 23: Maybe I'm missing something, but it seems like this class is currently not used under any circumstances. I guess the idea would be to enable it in case we're running it on XENON-home-turf in `alea_submission.py`, right?

hammannr (commented):
So far, I tested it on a non-XENON cluster and it worked, which is really nice! From reading the code, I would expect it to not fall back to the RCC defaults if run on midway. Maybe this part of the implementation is still missing. I can of course help with testing!

alea/submitters/rcc_slurm.py
kdund: correct, it is to keep any xenon functionality in this one

kdund:
Hi, @hammannr I checked off the two redundant imports, little idea why the build failure though-- opinions on merging? It is, in particular for XLZD use quite important to make alea non-midway-specific IMO :)

alea/submitters/rcc_slurm.py
hammannr: But what would I then do if I wanted to use this one on RCC? Right now, I can't select this in a config file, right?

kdund: ah added an rcc_slurm option to submission now :)

alea/scripts/alea_submission.py
```diff
@@ -34,6 +34,7 @@ def main():
     group = parser.add_mutually_exclusive_group(required=False)
     group.add_argument("--local", action="store_true", help="Executes the defined jobs locally")
     group.add_argument("--slurm", action="store_true", help="Prepare submission for slurm")
+    group.add_argument("--slurm", action="store_true", help="Prepare submission for slurm")
```
hammannr: I guess the second one should be `rcc_slurm`, right?

hammannr: I think apart from that should be good to go!

hammannr (approved):
I tested it with the modification I made and it worked. If you are also happy with this, please go ahead and merge. We'll fix the pre-commit thing separately.

PR #246, opened 2025-05-08
Parse the likelihood name to the base config

yuema137:
Currently, the sources don't know which likelihood they are in, as this information is not passed to the `pdf_base_config`
Adding this information is helpful in the cases when we want to apply specific treatment to each source when the likelihood names are different (for example sr0 and sr1)


GiovanniVolta (approved):
It sounds reasonable and the implementation is clean. I approve it

PR #265, opened 2026-02-24
Add argument `confidence_interval_root_find` indicating root finding method

dachengx:
When calculating CL, if `confidence_interval_root_find` is `brentq`, use `scipy.optimize.brentq`; if `extremal`, look for the left or right-most root.

The left- right- most root finding is performed in `utils.extremal_root`. This function accepts almost the same arguments as `scipy.optimize.brentq`.

### DIFF
--- a/alea/model.py
+++ b/alea/model.py
@@ -75,6 +75,7 @@ class StatisticalModel:
             threshold for confidence interval
         confidence_interval_root_find (str, optional (default="brentq")):
             root finding algorithm of confidence interval
+            supported options: "brentq" and "extremal"
         data (dict or list, optional (default=None)): pre-set data of the model
         fit_strategy (dict, optional (default=None)): strategy for the fit,
             see _DEFAULT_FIT_STRATEGY for possible settings
@@ -503,6 +504,7 @@ class StatisticalModel:
                 kind of confidence interval to compute
             confidence_interval_root_find (str, optional (default=None)):
                 root finding algorithm of confidence interval
+                supported options: "brentq" and "extremal"
 
         Returns:
             Tuple[str, Callable[[float], float], str, Tuple[float, float]]:
@@ -608,6 +610,7 @@ class StatisticalModel:
                 If None, the default kind of the model is used.
             confidence_interval_root_find (str, optional (default=None)):
                 root finding algorithm of confidence interval
+                supported options: "brentq" and "extremal"
             confidence_interval_args (dict, optional (default=None)): Parameters that will be fixed
                 in the profile likelihood computation. If None, all fittable parameters
                 will be profiled except the poi.
--- a/alea/model.py
+++ b/alea/model.py
@@ -12,7 +12,13 @@ from blueice.likelihood import _needs_data
 from inference_interface import toydata_to_file
 
 from alea.parameters import Parameters
-from alea.utils import within_limits, clip_limits, asymptotic_critical_value, ReadOnlyDict
+from alea.utils import (
+    within_limits,
+    clip_limits,
+    asymptotic_critical_value,
+    ReadOnlyDict,
+    extremal_root,
+)
 
 _DEFAULT_FIT_STRATEGY = {
     "disable_index_fitting": False,
@@ -67,6 +73,8 @@ class StatisticalModel:
             kind of confidence interval to compute
         confidence_interval_threshold (Callable[[float], float], optional (default=None)):
             threshold for confidence interval
+        confidence_interval_root_find (str, optional (default="brentq")):
+            root finding algorithm of confidence interval
         data (dict or list, optional (default=None)): pre-set data of the model
         fit_strategy (dict, optional (default=None)): strategy for the fit,
             see _DEFAULT_FIT_STRATEGY for possible settings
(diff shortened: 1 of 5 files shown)


hammannr:
On a first quick look this looks very nice! Could you think of a quick unites to make sure it stays that nice? 😄

dachengx:
Thanks @hammannr . I have added a simple test of the new method

tests/test_runner.py
```diff
@@ -38,6 +38,7 @@ def set_gaussian_runner(self, toydata_mode="generate_and_store"):
             nominal_values={"sigma": 1.0},
             parameter_definition=gaussian_model_parameter_definition,
             compute_confidence_interval=COMPUTE_CONFIDENCE_INTERVAL,
+            confidence_interval_root_find="extremal",
```
hammannr, line 41: Not really tests the method, but only if it fails to load it, no?

alea/utils.py
```diff
@@ -773,6 +774,65 @@ def correction_on_multiplier(x):
     return x
 
 
+def extremal_root(
+    f,
+    xL,
+    xR,
+    which="left",
+    step=0.01,
```
hammannr, line 783: Maybe we should assert that step is >=0, otherwise this will give unexpected behaviour or infinity loops.

alea/utils.py
```diff
@@ -773,6 +774,65 @@ def correction_on_multiplier(x):
...
+    step_growth=1.0,
```
hammannr, line 784: Same for step growth

hammannr: This option generally seems a bit unstable to me since it can explode or converge to zero, no?

alea/model.py
```diff
@@ -583,6 +606,8 @@ def confidence_interval(
             confidence_interval_kind (str, optional (default=None)):
                 kind of confidence interval to compute.
                 If None, the default kind of the model is used.
+            confidence_interval_root_find (str, optional (default=None)):
+                root finding algorithm of confidence interval
```
hammannr, line 612: For convenience, also state here the two options, please.

alea/utils.py
```diff
@@ -773,6 +774,65 @@ def correction_on_multiplier(x):
     return x
 
 
+def extremal_root(
```
hammannr, line 778: Currently, I see no way of adjusting these parameters in the config. Would it make sense to add this option with kwargs?

alea/utils.py
```diff
@@ -773,6 +774,65 @@ def correction_on_multiplier(x):
...
+    xR,
+    which="left",
+    step=0.01,
+    step_growth=1.0,
+    max_step=None,
+    xtol=1e-12,
+    rtol=4 * np.finfo(float).eps,
+):
+    """Find the left-most or right-most root of f in [xL, xR] using adaptive scanning + brentq
+    refinement."""
```
hammannr: Please add a brief explanation for all the parameters of the function.

hammannr (commented):
I didn't run it yet, but just had a look at the code. Looks mostly good, and I will try running it asap. In the meantime, I left a few comments on things I noticed so far.

tests/test_runner.py
dachengx: Here it is tested https://github.com/XENONnT/alea/blob/f9c6942b00626e931f7526bea28ecc9f5b60c8c2/tests/test_runner.py#L78

alea/utils.py
dachengx: Yes, it might explode, so I set the default `step_growth` as 1.0, and the step does not increase.

alea/utils.py
dachengx: I would not do it now. These arguments can be assigned in a notebook, but I do not want to add them now to the runners or submitters.

tests/test_runner.py
hammannr: sure but this doesnt' test if the function works properly, does it?

hammannr:
I played with it a little bit and it' quite easy to find a case in which this produces `RuntimeError: No root found in interval`:
This works fine:
```python
def function_with_six_roots(x):
    return (x - 1) * (x - 2) * (x - 3) * (x - 4) * (x - 5)* (x - 6)

left_extremal = extremal_root(
    function_with_six_roots,
    xL=-100,
    xR=100,
    which="left")
```
though this will fail without readjusting the step size:
```python
def function_with_six_small_roots(x):
    return (x - 1e-5) * (x - 2e-5) * (x - 3e-5) * (x - 4e-5) * (x - 5e-5)* (x - 6e-5)

left_extremal = extremal_root(
    function_with_six_small_roots,
    xL=-100,
    xR=100,
    which="left",)
```
Do you think there is a sensible way to make this a bit more robust?

dachengx:
> I played with it a little bit and it' quite easy to find a case in which this produces `RuntimeError: No root found in interval`: This works fine:
> 
> ```python
> def function_with_six_roots(x):
>     return (x - 1) * (x - 2) * (x - 3) * (x - 4) * (x - 5)* (x - 6)
> 
> left_extremal = extremal_root(
>     function_with_six_roots,
>     xL=-100,
>     xR=100,
>     which="left")
> ```
> 
> though this will fail without readjusting the step size:
> 
> ```python
> def function_with_six_small_roots(x):
>     return (x - 1e-5) * (x - 2e-5) * (x - 3e-5) * (x - 4e-5) * (x - 5e-5)* (x - 6e-5)
> 
> left_extremal = extremal_root(
>     function_with_six_small_roots,
>     xL=-100,
>     xR=100,
>     which="left",)
> ```
> 
> Do you think there is a sensible way to make this a bit more robust?

Maybe 1% of the parameter's boundary?

hammannr:
> Maybe 1% of the parameter's boundary?
This could work but sometimes people (including myself) use relatively large parameter bounds to ensure the limits are well within. So also there would be some source of stability. This seems like it should be a solved problem, though.

dachengx:
> > Maybe 1% of the parameter's boundary?
> > This could work but sometimes people (including myself) use relatively large parameter bounds to ensure the limits are well within. So also there would be some source of stability. This seems like it should be a solved problem, though.

I hope the analysts will figure out how to scale the parameter of interest to be within meaningful values. I will print a warning when the step is too large/small compared to the poi's boundary.

PR #270, opened 2026-04-16
All sorted source names

GiovanniVolta:
Renamed property to reflect that source names are sorted alphabetically.

Check https://xenonnt.slack.com/archives/C07ML1P7YH1/p1776351252147869

### DIFF
--- a/alea/models/blueice_extended_model.py
+++ b/alea/models/blueice_extended_model.py
@@ -209,7 +209,7 @@ class BlueiceExtendedModel(StatisticalModel):
             # sum over sources with same names of all likelihood terms
             ret = {
                 n: sum([ret[ll_name].get(n, 0.0) for ll_name in ret.keys()])  # type: ignore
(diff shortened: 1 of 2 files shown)


hammannr:
@GiovanniVolta thanks for bringing this issue up. I'm hesitant to change names of properties though for backward-compatibility. In #232 I improved the docstrings to make the distinction clearer; they now read:
```python
def get_source_name_list(self, likelihood_name: str) -> list:
        """Return a list of source names for a given likelihood term.

        The order matches the ``source`` column of the data, so this can be used to map
        indices in the data to a source name.

        ...
        """

def all_source_names(self) -> list:
        """Return a sorted list of all source names across all likelihood terms."""
```

Do you think this is enough to prevent future confusion? In that case, I'd close this PR.

GiovanniVolta:
Yes, it looks fine for me. I will close the PR

